PR #9, opened 2016-08-14
display update notifications when a newer ghost cli is available

### LINKED ISSUE #6: [FR] update notifier
It would be nice if the package realizes when it's outdated and notifies the user.
This proofed to be useful for other cli projects in the past.

We've used:
https://github.com/yeoman/update-notifier in the past.
It's pretty easy to use and never resulted in problems until now.

sakulstra:
- closes #6

### DIFF
--- a/bin/ghost
+++ b/bin/ghost
@@ -4,6 +4,6 @@
 // provide a title to the process
 process.title = 'ghost';
 
+require('../lib/utils/update-notifier');
 var cli = require('../lib/cli');
-
 cli(process.argv);
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
     "npm": "2.15.9",
     "pm2": "1.1.3",
     "request": "2.74.0",
-    "semver": "5.3.0"
+    "semver": "5.3.0",
+    "update-notifier": "1.0.2"
   },
   "devDependencies": {
     "chai": "3.5.0",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -48,7 +48,8 @@
     "npm": "2.15.9",
     "pm2": "1.1.3",
     "request": "2.74.0",
-    "semver": "5.3.0"
+    "semver": "5.3.0",
+    "update-notifier": "^1.0.2"
```
acburdine: Could you lock this version to a specific one please? 😄

bin/ghost
```diff
@@ -4,6 +4,6 @@
 // provide a title to the process
 process.title = 'ghost';
 
+require('../lib/update_notifier');
```
acburdine: This is being a bit nitpicky, but could you move the `update_notifier` into the `utils` folder/make it `update-notifier` instead? That would keep it a bit more consistent with the file naming schema 😄

bin/ghost
```diff
@@ -4,6 +4,6 @@
 // provide a title to the process
 process.title = 'ghost';
 
-var cli = require('../lib/cli');
-
+require('../lib/utils/update-notifier');
+var cli = require('../lib/utils/cli');
```
acburdine: this will break the application as the `cli` file is not in the utils folder - it can be left out of the utils folder since it's listed as the "main" file in package.json

PR #44, opened 2016-09-27
Add url as parameter

MikaXII:
Can use ghost config --url <value> as parameter.


acburdine:
Hey @MikaXII - I will try and get to this today. I'm fairly certain this fix works but need to verify :)

acburdine:
Thanks @MikaXII!

PR #57, opened 2016-10-13
✨  use knex-migrator

kirrg001:
### TODO
- [x] test this code
- [x] knex-migrator package.json
- [x] forward path to knex migrator where to find the .knex-migrator file

Check database health before starting Ghost and run `knex-migrator` init if database was never initialised.

### DIFF
--- a/lib/commands/start.js
+++ b/lib/commands/start.js
@@ -1,5 +1,4 @@
-var KnexMigrator = require('knex-migrator'),
-    BaseCommand = require('./base');
+var BaseCommand = require('./base');
 
 module.exports = BaseCommand.extend({
     name: 'start',
@@ -29,7 +28,7 @@ module.exports = BaseCommand.extend({
             Promise = require('bluebird'),
             path = require('path'),
             self = this,
-            environment, config, pm, cliConfig, knexMigrator;
+            environment, config, pm, cliConfig;
 
         options = options || {};
         environment = (options.production || !options.development) ? 'production' : 'development';
@@ -46,6 +45,8 @@ module.exports = BaseCommand.extend({
         // TODO: rethink this
         return this.runCommand('config', 'paths.contentPath', path.join(process.cwd(), 'content'), {environment: environment})
             .then(function () {
+                var KnexMigrator = require('knex-migrator'),
+                    knexMigrator;
                 /**
                  * If we require any JS file from Ghost source code (for example the config), then it's always env-less!
                  *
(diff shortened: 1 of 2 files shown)


lib/commands/start.js
```diff
@@ -45,6 +46,13 @@ module.exports = BaseCommand.extend({
         // TODO: rethink this
         return this.runCommand('config', 'paths.contentPath', path.join(process.cwd(), 'content'), {environment: environment})
             .then(function () {
+                knexMigrator = new KnexMigrator();
```
acburdine: I know this is still a WIP - but in order to improve load times of the CLI, there's a better way to have the requires work:

```js
var KnexMigrator = require('knex-migrator'),
    knexMigrator = new KnexMigrator();

// ... do other things
```

If you do this and remove the two changes above this then the code changes look good to me - will test it locally once it's out of WIP status :smile:

kirrg001:
@acburdine the new Ghost-CLI version won't run with an older Ghost alpha version when merging this PR

acburdine:
@kirrg001 I went ahead and changed the code to what I was asking (because needing to get this merged) - but here's why:

The cli has to require each command file (so it can get all of the necessary variables like command name, arguments, options, etc). Because of this, adding a bunch of require statements to the top of the file increases load times, especially when the command is not even being run (for example, if you run `ghost config` - you don't even need `ghost start` or its dependency on "knex-migrator", yet with your code knex-migrator would still be loaded).

Basically, this is probably over-ambitious optimization, but as it was a quick change and the same practice is followed everywhere else in the code, I think it's necessary.

I'm willing to revert this change in the future if there's good reason to, though 😄

lib/commands/start.js
acburdine: see below comment - I see why you didn't instantiate knex-migrator until later on in the code, but at the same time moving the variables down makes more perf sense imo

acburdine:
So I pulled this locally to test - and it's not working - neither sqlite nor mysql dependencies are installed, so ghost start doesn't work. And I'm not sure that having ghost-cli depend on both sqlite and mysql is a good idea... :confused:

kirrg001:
> So I pulled this locally to test - and it's not working - neither sqlite nor mysql dependencies are installed, so ghost start doesn't work. And I'm not sure that having ghost-cli depend on both sqlite and mysql is a good idea... 😕

`ghost start` might not work because of a missing PR in Ghost?
https://github.com/TryGhost/Ghost/pull/7585/files

PR #60, opened 2016-10-17
🐛  ui shows [object object] when error is not instance of Error

kirrg001:
no issue


acburdine:
This LGTM, but do you know where the error was being thrown from? It would be good to know where it's coming from so I know where to add better error handling :)

kirrg001:
Yes it was thrown in Ghost and forwarded from `knex-migrator` into Ghost-CLI, will check soon.

kirrg001:
https://github.com/TryGhost/Ghost/pull/7582

PR #68, opened 2016-10-20
Convert JSCS/JSHint => ESLint

acburdine:
- eslint is becoming the community standard for linting javascript (jscs has been deprecated/merged into eslint)
- lint checks are only run on node v4 (which is eslint's minimum supported version), so this will work


acburdine:
@ErisDS I used [this tool](https://github.com/brenolf/polyjuice) to convert the jscs/jshint rules to an eslint configuration. Required a bit of modification after that, but would there be interesting in porting these changes to Ghost itself? The initial .jscs/.jshint files are direct copies of Ghost's configuration.

PR #79, opened 2016-11-03
🎨 Minor messaging improvements

ErisDS:
- Output version information when install/downloading
- Include a default when prompting for URL

Note, locally when running the tests, even on master I get a failure:

```
  1) Unit: npm correctly captures the npm output and returns it when output is a string:
     TypeError: this is not a typed array.
      at Function.from (native)
      at Context.<anonymous> (test/unit/utils/npm-spec.js:155:43)
```


acburdine:
LGTM 👍

PR #82, opened 2016-11-07
:art: make npm utility use spawnSync

acburdine:
no issue
- makes npm utility more concise as it now relies upon spawnSync's built-in behavior for capturing output


acburdine:
As it turns out - this actually doesn't work correctly because it's technically a synchronous call :/ will revisit later potentially.

PR #130, opened 2017-01-24
🎨  knex-migrator: migrate if needed

kirrg001:
no issue

**requires knex-migrator: 2.0.1 , see #131**

- if knex-migrator tells ghost-cli that migrations need's to be executed, we care about this
- problem description:
  - if you upgrade Ghost to a higher version and the new version contain migration scripts
  - we need to ensure that ghost run/start handles that for the user
  - if we don't ghost-cli won't start and the user is confused


lib/commands/run.js
```diff
@@ -40,9 +40,10 @@ module.exports.Command = BaseCommand.extend({
         return knexMigrator.isDatabaseOK()
             .catch(function (error) {
                 if (error.code === 'DB_NOT_INITIALISED' ||
-                    error.code === 'DATABASE_DOES_NOT_EXIST' ||
```
kirrg001: This code does not exist. (anymore?) - deleted it

acburdine: 👍 It existed when knex-migrator was built first, was likely removed later.

acburdine:
Test failures are unrelated to this PR - should be fixed once it's merged.

PR #159, opened 2017-02-19
[WIP] Nginx

acburdine:
TODOS:

- [ ] test this on a live server with a url

The initial version of the Nginx service will _not_ support:
- subdirectories
- existing config files

While the library I'm going to use `nginx-conf` should make integrating the above much easier, there are some edge cases that we may want to test for, which we can wait to do until after 1.0 ships.


acburdine:
Closing in favor of a different approach.

PR #172, opened 2017-03-16
Cut down trees (or in other words, add logging)

### LINKED ISSUE #42: Improve logging output
Right now ghost-cli starts development mode and production mode in background by spawning a child process.

I am working on logging for ghost and i wonder how this can look like with ghost-cli.
We can create a command like `ghost log`, which is aware of where the log rotated files are and makes a tail of the newest file. Not sure about development mode, because usually we log in development to stdout/stderr only. That would mean `ghost start name=development` would not start the process in background and pipes the stdout/stderr to current process.

acburdine:
closes #42 

- adds logging config command line option
- adds `ghost log` command to be able to parse & view the bunyan logs of a running Ghost process (and also follow them 😏 )

~~Currently blocked on some issues with Ignition's [PrettyStream class](https://github.com/TryGhost/Ignition/blob/master/lib/logging/PrettyStream.js). (Open PR: TryGhost/Ignition#35)~~


acburdine:
Travis is having issues atm, going to preemptively merge this as the tests have passed, it's the lint step that hasn't and lint succeeds locally.

Issue #194, opened 2017-05-03
Programmatic use of Ghost-CLI


ErisDS:
Ghost-CLI's interactive UI is super-crazy-awesome for users in the most common use-cases, which is doing a 1 time install on either a new or existing server.

However, in certain cases, it's desirable to be able to call ghost-cli programmatically, rather than manually & interactively. Ordinarily, I'd say it was up to the consumer to figure out how to do this, however the way that ghost-cli's interactive UI works by rewriting the prompt makes it very hard to answer the prompts programatically with something like [expect](https://linux.die.net/man/1/expect).

Therefore, I'd like to introduce a path-of-least-resistance workaround, which is to make it possible to pass in enough arguments that ghost-cli knows what you want it to do and doesn't output any further prompts.

To achieve this with the current features of Ghost-Cli I propose two additional arguments for the CLI install/setup commands:

- `sslemail`, e.g. `--sslemail=me@mine.com`, toggle SSL setup with this address
- `start`, e.g. `--start` or `--start=false`, toggles whether or not Ghost should be started

acburdine:
Going to close this for now as the two merged PRs have fixed this issue temporarily - Further improvements upon this will happen as a part of #179

PR #195, opened 2017-05-03
✨ Allow SSL Email to be provided, skip prompts

ErisDS:
This is a proposal for how to introduce ssl config without requiring a prompt.
It is a little bit odd in that it causes the ssl email to be written to config. Not sure if that's useful/desired.

Note: I've not done particularly extensive testing of this.

refs #194

- If --sslemail is provided, use that instead of prompting
- Makes it easier to call ghost programmatically

### DIFF
--- a/lib/commands/config/advanced.js
+++ b/lib/commands/config/advanced.js
@@ -84,6 +84,10 @@ module.exports = [{
     description: 'Database name',
     configPath: 'database.connection.database'
 }, {
+    name: 'sslemail',
+    description: 'SSL email address',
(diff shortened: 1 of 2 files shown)


acburdine:
I'll pull this down and test tomorrow

ErisDS:
Can we get this, or something that achieves this into a release please? From middle of next week I will be blocked on #194 - the concept of programatically calling ghost-cli 😬

acburdine:
Yup - that will be my focus this weekend

Issue #179, opened 2017-03-20
Move away from commander.js for handling command line args & options


acburdine:
When Ghost-CLI was first designed, commander.js was used to handle args & options because it was one of the more popular solutions out there, and is fully featured.

However, as the CLI has evolved, certain requirements that commander.js enforces make it no longer a useful/viable solution for handling command line params. The main one is that commander must know about _all_ possible arguments and options prior to running the command, leaving little to no room for custom options added by other commands, or by [services](https://github.com/TryGhost/Ghost-CLI/issues/146) (for example, the strictness of this requires lines like [this one](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands.js#L51)).

It also requires parsing args and options config and then converting it to a string, which then gets converted back into options in commander.js (see [this code](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/cli.js#L11-L63)).

The one big pro that commander.js _does_ provide is automatic generation of help text for commands like `ghost help install` or `ghost install --help`. However, it would be trivial to manually  implement this in the CLI itself

The two big alternatives to commander that I've seen used in a lot of places are:
- [minimist](https://github.com/substack/minimist) - This provides very little in the way of automatic enforcement of args/options, but allows for the greatest configurability
- [yargs](https://github.com/yargs/yargs) - This provides the same features as minimist re: basic arg parsing, and also allows for greater enforcement of correct args if wanted. This library also seems to be more active than either commander.js or minimist

PR #196, opened 2017-05-03
✨ Allow start flag to be provided, skip prompts

ErisDS:
Again, this is a proposal more than anything & not particularly well tested 😁 

refs #194

- if --start is provided, use that instead of prompting
- makes it easier to call ghost install/ghost setup programmatically


acburdine:
Because `ghost install` calls the `ghost setup` command automatically, this change will only work if you run `ghost setup` as a standalone command. This is due to limitations in commander.js and how it handles options, and could potentially be resolved by #179

PR #199, opened 2017-05-05
fix(run): make command descriptions clearer

### LINKED ISSUE #185: Clearer explanation for `ghost start` vs `ghost run`
```
    restart                         Restart the Ghost instance
    run                             start a managed ghost process
    service <command> [args...]     run a service-defined command
    setup [options]                 setup an installation of Ghost (after it is installed)
    start                           starts an instance of Ghost
    stop [options]                  stops a named instance of Ghost
```

The difference between run and start isn't very clear. Although `start`/`stop`/`restart` follows the normal pattern, so it's reasonably obvious which one I should use, but I almost think `run` doesn't need help output?

Surely the only people who ever need `run` are people developing process managers?

acburdine:
closes #185
- fix command descriptions
- output help note on `ghost run` if it is being run directly

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -10,6 +10,13 @@ let instance;
 module.exports.execute = function execute() {
     checkValidInstall('run');
 
+    // If the user is running this command directly, output a little note
+    // telling them they're likely looking for `ghost start`
+    if (process.stdin.isTTY) {
+        this.ui.log('The `ghost run` command is used by the configured Ghost process manager and for debugging. ' +
+            'If you\'re not running this to debug something, you should run `ghost start` instead.', 'yellow');
+    }
+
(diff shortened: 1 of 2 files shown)


lib/commands/run.js
```diff
@@ -10,6 +10,13 @@ let instance;
 module.exports.execute = function execute() {
     checkValidInstall('run');
 
+    // If the user is running this command directly, output a little note
+    // telling them they're likely looking for `ghost start`
+    if (process.stdin.isTTY) {
+        this.ui.log('The `ghost run` command is used by the configured Ghost process manager and for debugging. ' +
```
acburdine: /cc @ErisDS does this message make more sense?

ErisDS: 👍

Issue #190, opened 2017-05-02
SSL handling


acburdine:
This issue stems from comments in #188, and will hopefully help address #176.

**One of the most likely issue that users are going to run into is trying to set up SSL _before_ they have adequately configured their domain's DNS record to point to the current server's IP.**

This DNS setup is a *required* pre-requisite of setting up ssl for ghost, which we can write in documentation; however, ghost-cli should first do a bit more checking to prevent confusing errors.

Before running the actual letsencrypt cert provisioning command, the CLI should do a DNS check (using node's built in dns module) to ensure proper setup. If a proper DNS record is not found, the CLI should exit with a message telling the user to first setup the DNS entry correctly, then run the `ghost nginx-ssl` command to finish setting up SSL. The CLI should finish any additional setup after presenting that error.

**We need to prompt the ssl handling, see #177.**
**Ensure we tell the user that we are going to generate a letsencrypt for him**

acburdine:
An initial version of this has been completed in #259, however there will be a more complex (but better and more thorough) solution introduced after 1.0.0

acburdine:
With #507, there were some additional checks implemented that handles the "ip is pointing to the wrong server IP" error.

Going to close this now, we can reopen in the future if we decide we need better checks.

arvindvenkataramani:
My domain is correctly setup (A, CNAME) and I'm still getting this error:

Message: Your domain name is not pointing to the correct IP address of your server, please update it and run `ghost setup ssl` again

argais:
Issue persists, ghost cli 1.11.0, dns has been set for hours, and is resolving correctly to the IP it should be and still failing with the same message...

Issue #189, opened 2017-05-02
Ghost user creation (for systemd)


acburdine:
One of the current weird parts of the ghost-cli setup is its needed privileges, both for setting up a new Ghost blog (via the CLI itself) and for actually _running_ the Ghost node process. To ensure transparency as well as security, there needs to be a bit of additional handling done in the CLI where system users are concerned. The first part of this is to ensure that whoever actually runs `ghost install` knows what things the CLI needs to run as sudo. There is an issue open currently to track this: #164. 

However, the part this issue covers is configuring and then running the actual persistent node process as a user without any type of sudo permissions, made simpler thanks to systemd, which allows the ghost process to be run under another user. So, what Ghost-CLI in the end needs to do is create a new user on setup (similar to what a mysql installation does), and then run Ghost as that user via the configured systemd service. This user would be named `ghost`, and have the login shell set to `/bin/false` to prevent actually logging in as that user. For multiple blogs setup on one system, the same ghost user would be re-used for both of them (but could potentially be changed via a command line option?). The user would also need to be made owner of the ghost install directory.

The only really tricky bit with this arises when one-click/automatic updates are implemented in the future. The way the current, manual `ghost update` command works, the new version is downloaded, dependencies are installed, and then ghost is stopped & restarted using `ghost stop` and `ghost start`, both of which require `sudo` due to the interaction with systemd. However, the actual download & dependency installation steps don't require sudo, so that should be possible with the non-sudo `ghost` user, which would need to be done with automatic updates since the update request would be triggered from the running Ghost process. During said automatic update, the actual ghost process would kill itself after installing the updated version, relying on systemd to restart the node process with the new updated version. (see TryGhost/Ghost#8005) for a little bit more context on automatic/one-click updates).

**We need to prompt the user creation. See #177.**

--

A user without sudo access won't be able to use systemd regardless. Will fallback to local process manager. So I think what really needs to happen is just better documentation on how to _not_ use systemd e.g. ghost install --process local

Systemd will restart Ghost if ghost stops for whatever reason - it also will restart Ghost on server reboot - whereas the local one won't

Local process manager can't be run as a separate user either - you have to run ghost start as the user you intend to run Ghost with

e.g. if you run `ghost install` and we prompt you to create the user `ghost`, but it doesn't work we fallback to local process manager.

PR #200, opened 2017-05-06
I am become uninstall, destroyer of ghost instances

### LINKED ISSUE #187: `ghost uninstall` command
With the introduction of services, the `nginx` and `systemd` services now add files elsewhere in the system that complicate uninstalling a blog (meaning you can't just `rm -rf /your/ghost/blog`).

Ghost uninstall will first stop everything, then call each service that needs to cleanup something, then delete everything from the directory it's in.

Also, `ghost uninstall` will ask if the user is super-duper mega extra sure if they want to remove their blog because there's no going back afterwards.

The only thing that won't be cleaned up after an installation is the database if it's mysql. In theory we could use `knex-migrator reset` to do it, but I'm not sure that's really necessary? People might want to delete a blog and keep all the data.

acburdine:
~~Almost done - just need to test on Linux and also add a check for the nginx service so that it doesn't try to delete files that aren't there.~~ done!

closes #187

### DIFF
--- a/lib/commands.js
+++ b/lib/commands.js
@@ -120,7 +120,7 @@ module.exports = {
     },
 
     update: {
-        description: 'update a Ghost instance',
+        description: 'Update a Ghost instance',
 
         arguments: [
             {name: 'version', optional: true}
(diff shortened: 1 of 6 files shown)


acburdine:
NOTE: there is currently a bug with this command because the "root" directory in the ghost install directory has a root ownership (if you create an SSL certificate). While it _could_ be fixed as part of this issue, I will solve it instead as a part of the larger changes in #189 and #190

PR #213, opened 2017-06-18
feat(ui): show sudo command before it is run

### LINKED ISSUE #164: Always inform user when sudo is required
Because of the nature of some of the setups (systemd, nginx configuration), sometimes a sudo-level command must be run by the CLI. To be as transparent and clear (e.g. not confusing) as possible, any CLI action that requires sudo access should first log and inform the user as to what is being done.

If verbose output is not turned on, just a basic description should be logged, e.g. "Systemd configuration is being moved into place, you may be prompted for your password". However, if verbose logging is enabled, the actual command should be logged as well.

acburdine:
closes #164
- add sudo ui method that shows the sudo command before it is run
- make log method synchronous again

### DIFF
--- a/lib/ui.js
+++ b/lib/ui.js
@@ -3,6 +3,7 @@ const fs        = require('fs-extra');
 const path      = require('path');
 const ora       = require('ora');
 const chalk     = require('chalk');
+const execa     = require('execa');
 const errors    = require('./errors');
 const Table     = require('cli-table2');
 const assign    = require('lodash/assign');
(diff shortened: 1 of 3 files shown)


lib/ui.js
```diff
@@ -97,7 +113,7 @@ class UI {
             message = chalk[color](message);
         }
 
-        this.noSpin(() => this.stdout.write(`${message}\n`));
```
acburdine: 🤦‍♂️ surprise asynchronous code in a synchronous method

PR #222, opened 2017-06-22
feat(cli): swap out commander for yargs parsing

acburdine:
refs #179 [ci skip]
- make Command class an actual class
- refactor bootstrap & clean up command invocation stack

TODOS:
- [x] finish refactoring commands
- [x] diagnose issue with killing ghost run (something with the cleanup method)
- [x] fix the tests
- [x] little bit of spit & polish (code cleanup)
- [x] 💸


acburdine:
The fact that removing one of the test files only dropped coverage by 3% just goes to show how bad the test coverage is right now 😭

Issue #248, opened 2017-06-28
Systemd enable/disable behaviour


ErisDS:
In order to fully setup Ghost with systemd, we need to call `systemctl enable <servicename>`.

Short term, we're going to do this as part of setup, just to get the call in to the codebase.

Longer term, we need a slightly more elegant solution.

The problem with enabling during setup, is that if the user chooses not to start Ghost, it will be started if the server is then restarted. This is a bit odd.

Instead, we need some more elegant logic to handle calling enable on the first start, if enable has not already been called.

The suggested logic is:

- systemd could add something like an `--enable` option to `ghost start` and a `--disable` option to `ghost stop`
- systemd would check if it's enabled on `ghost start`
- if it's not and `--enable` is _not_ passed, prompt the user and ask if they want to enable
- if it is enabled do nothing regardless of whether or not `--enable` is passed (edited)
- for stop, it wouldn't disable unless `--disable` is passed (edited)
- the `stop --disable` part is for people who are stopping a blog but _not_ uninstalling it

Issue #216, opened 2017-06-20
Tasks before going live (error messages, documentation, improvements)


cobbspur:
## Documentation at docs.ghost.org (start with this on the 7th of July - not before please)

We are going to have a little call about the CLI documentation on the 7th of July. That's why the sub tasks here are unassigned.

@cobbspur 
- [x]  differ between local and production installation ( ghost install --development ghost install local etc.)
- [x] document database options e.g. mysql vs. sqlite (mysql user creation sudo permissions) - how to install Ghost with sqlite? e.g. ghost install --url https://kate-ghost-test.acburdine.me --db sqlite3 --dbpath ./content/data/ghost.db
- [x] CLI prompts nginx installation/template creation -> document this and document that we are using letsencrypt. letsencrypt is accepted by browsers and is free ssl cert generation
- [x] document #211
- [x] write a little section about where the nginx config file lives and how to modify this file
- [x] write a little section about where the ssl config file lives and how to modify this file

@AileenCGN 
- [x] document `ghost doctor`
- [x] skip nginx/ssl installation via command line e.g. `--nosetup`, --no-setup-ssl, --no-start - document special flags (see also ghost help)
- [x] more documentation about `ghost config`
- [x] document verbose flag
- [x] document ghost setup [service] e.g. ghost setup ssl (install extensions manually/separately/afterwards)

@ErisDS 
- [x] document how to use the CLI in a provision script - see conversation https://ghost.slack.com/archives/C04CUS2R3/p1498770776174257

@kirrg001
- [x] environment requirements (minimum memory required), ubuntu version >= 16.X?
- [x] never install ghost inside /root - add a 100% recommended way e.g. install node globally /usr/local/node and /usr/local/nvm and create a folder in /var/www/
- [x] ghost-cli needs root access, if you have a custom ubuntu user, it needs to be in the sudoers group
- [x] document the default permission requirements for the default CLI installation (=== sudo access)
- [x] document that systemd setup is REQUIRED (otherwise use --process local)
- [x] if `ghost start` is somehow not working, you can run `ghost run` which might help you to clarify to issue you are running in (document ghost run as best as we can)
- [x] document that if you skip systemd, you will get "Message: Systemd process manager has not been set up. Run `ghost setup linux-user systemd` and try again." - that means you have to set it up. ONLY if systemd can't be used, we auto fallback (user can use --process local to force local process manager)
[...]

PR #249, opened 2017-06-28
✨ feat(setup): enable systemd service during setup

ErisDS:
This is a quick version for now, to ensure that enable is called and Ghost doesn't fail to start on reboot of servers.

I will try to revisit the better plan in #248 next week,

refs #216, #248

- add an extra step to enable systemd on setup
   - this is what tells systemd to start the service on server restart
- Update the template to contain docs link

### DIFF
--- a/lib/services/process/systemd/ghost.service.template
+++ b/lib/services/process/systemd/ghost.service.template
@@ -1,6 +1,6 @@
 [Unit]
 Description=Ghost systemd service for blog: <%= name %>
-# Documentation=add documentation link here
+Documentation=https://docs.ghost.org
 
 [Service]
 Type=simple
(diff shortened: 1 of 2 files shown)


lib/services/process/systemd/ghost.service.template
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Ghost systemd service for blog: <%= name %>
-# Documentation=add documentation link here
+Documentation=https://docs.ghost.org
```
rv-aburdine: 🙈

acburdine:
@ErisDS could I get you to rebase this against master? I won't be able to do it myself 😛

acburdine:
sooo...kind of stole the thunder on this as I wanted to get the full feature of enable/disable in. Would it be possible to update this PR/message just to update the docs link in the template?

ErisDS:
Doesn't really matter how it gets in :) You are definitely more well-placed to do a rebase on this than I - I just tried and I don't know where any files are anymore and there are no instructions or explanations that I can find, so I'm going to have to relearn the codebase



It's totally possible for you to rebase. If the only thing left to do is a text change then might as well just commit that directly?

PR #259, opened 2017-07-03
Nginx Improvements

acburdine:
Two things in this PR:
- feat(nginx): check dns points to ip before attempting ssl setup
  refs #190
  - adds check to ssl setup that ensures DNS has been properly configured before trying to get an SSL certificate

- ssl renew command for nginx extension

TODOS
- [ ] test on droplet
- [ ] add cron script (may come in separate PR)


acburdine:
...and this works 🕺

PR #263, opened 2017-07-04
fix(systemd): ensure systemd service file is created before starting

### LINKED ISSUE #262: Ghost doesn't start when nginx setup and systemd not
When setting up Ghost via the CLI and I chose to configure nginx and say 'no' when promted for systemd, Ghost won't start:

image

OS: Ubuntu 16
Ghost-CLI: 1.0.0-beta.3

acburdine:
closes #262
- add precheck to all systemd commands todetermine whether or not systemd is set up before running

### DIFF
--- a/extensions/systemd/systemd.js
+++ b/extensions/systemd/systemd.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const fs = require('fs');
 const execa = require('execa');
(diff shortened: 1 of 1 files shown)


acburdine:
Tested on ubuntu and it works 👍

Issue #191, opened 2017-05-02
Mysql User Creation


acburdine:
To simplify and secure the mysql handling in the CLI and in Ghost itself, it is a good idea to configure Ghost with a mysql user that only has privileges for the exact database that Ghost requires. 

What Ghost-CLI will need to do is prompt (see #177) for the mysql _root_ username and password as well as a database to use for Ghost itself (it already does this), but after doing that run raw mysql commands to create that database and another database with a `_backup` suffix (to be used as a part of #153), then create a `ghost` mysql user that only has permissions for those two databases on localhost. The password for this ghost user will be an auto-generated random alphanumeric string (it will be put in `config-<env>.json` so if the user wants to know what the password is, it will be there).

This should hopefully ensure a good, secure solution for the mysql connection.

Questions:
- what if the user doesn't have the root credentials?
- what if people want to use their own user name -> probably just detect if user name is root and continue?

Later:
- we won't implement the backup feature itself, we just create the table

ErisDS:
FYI - I sorta ran into this again. 

Not sure how readable this is - but this is my strategy for getting from Ubuntu 16 to running Ghost-CLI: https://gist.github.com/ErisDS/0edb9c2bca72f99b6a428ae9f17cb747

There are other steps for security, but these are the most relevant ones to ghost-cli.

Note that I create a DB and mysql user for ghost and pass these deets in.

Note also that I create a user for ghost that I'm not really using yet, cos I'm not sure what I should do here / vs what the cli is going to do, which is more relevant to #189.

PaszaVonPomiot:
I've read through yml it would be good to have an option to enter webroot path for ghost instance and database name because if someone would like to have multiple instances he would probably go for something like /var/www/domain1-com, /var/www/domain2-com, ...

kirrg001:
We won't create the backup table now. We can also do that later.

The MySQL extension will do the following:
- it first checks if it can connect to the MySQL database with the entered credentials from the configuration step
- it checks if the username is root and if so, we prompt the user if the CLI should create a custom ghost mysql user with a random, strong password (this user has a limited access to the ghost database)
- if the username is not root, we just proceed with the installation
- we ensure that the config get's updated if we create a custom mysql user

PR #269, opened 2017-07-04
Mysql extension

kirrg001:
refs #191 

### to discuss/edge cases
- If the mysql `ghost` user exists, the CLI can detect that, but that's it - we won't be able to figure out the password - so imagine you run `ghost setup` twice. on the first run the CLI creates the mysql user and stores this user inside the config.production.json. Then the user manually deletes this file. And he runs `ghost setup` again, he enters the root credentials again. The CLI would detect that the custom mysql user exists, but without a password, we can't do anything. Edge case?
- there is a tricky case: imagine we have created the mysql user for you. the ghost user is part of the config file. alright, now you re-run `ghost setup` and change the database name. You can't change your privileges without the root credentials, we would have to prompt the credentials again.

### TODO
- [x] remove hotfix for lib/commands/doctor/checks/setup.js, wait that the fix is merged

### DIFF
--- a/extensions/mysql/index.js
+++ b/extensions/mysql/index.js
@@ -2,6 +2,7 @@
 
 const Promise = require('bluebird');
 const mysql = require('mysql');
+const crypto = require('crypto');
 const omit = require('lodash/omit');
 const cli = require('../../lib');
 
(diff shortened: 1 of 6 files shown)


extensions/mysql/index.js
```diff
@@ -0,0 +1,118 @@
...
+        // ghost setup --local, skip
+        if (argv.local) {
+            return;
+        }
+
+        cmd.addStage('mysql', this.setupMySQL.bind(this));
+    }
+
+    setupMySQL(argv, ctx, task) {
+        let self = this;
```
acburdine: Just so we keep in line with the ES6 in the rest of the CLI codebase, can we replace all of the `let self = this;` with arrow functions please?

extensions/mysql/index.js
```diff
@@ -0,0 +1,118 @@
+'use strict';
+
+const Promise = require('bluebird');
+const mysql = require('mysql');
```
acburdine: the CLI doesn't depend on `mysql` explictly. Although it's added indirectly by knex-migrator, we need to add this dep to Ghost-CLI itself if we want to ensure things to work right in all installs.

extensions/mysql/index.js
```diff
@@ -0,0 +1,118 @@
...
+            });
+        });
+    }
+
+    createMySQLUser(ctx) {
+        let self = this;
+        let randomPassword = crypto.randomBytes(10).toString('hex');
+        let host = self.databaseConfig.connection.host;
+
+        return new Promise(function (resolve, reject) {
```
acburdine: Instead of the `return new Promise((resolve, reject) => {})` usage - bluebird has a `Promise.fromCallback` method, would probably be good to abstract that out into a helper method in this class.

e.g.
```js
_query(queryString) {
  return Promise.fromCallback(cb => this.connection.query(queryString, cb));
}
```

extensions/mysql/index.js
kirrg001: 👍

extensions/mysql/index.js
kirrg001: Yeah we can do that 👍

extensions/mysql/index.js
kirrg001: 👍

lib/errors.js
```diff
@@ -51,6 +51,14 @@ class CliError extends Error {
     toString(verbose) {
         let output = `${chalk.yellow('Message:')} ${this.message}\n`;
 
+	if (this.options.context) {
```
acburdine: :troll: needs indent fix

package.json
```diff
@@ -57,6 +57,7 @@
     "knex-migrator": "2.0.16",
     "listr": "0.12.0",
     "lodash": "4.17.4",
+    "mysql": "^2.13.0",
```
acburdine: can this be locked down to a specific version like all our others are?

kirrg001:
Ready for review 🚓

acburdine:
Merging this for now, will clean up afterwards

PR #274, opened 2017-07-05
🐷  Improve config error logs and messages

AileenCGN:
refs #216

This PR changes the `Config.exists()` method in a way that rather than simply returning false, we're catching the error and return a `ConfigError` object that includes the received error object from `fs.readJsonSync()` as a message.

There are two places, where `Config.exists()` is being called and was expecting a boolean `false` to determine if the config is valid. But this can mean, that the config either doesn't exist or has a syntax error. It remained a secret to the user.

I changed this to check for an `options` property (I'm not very happy with the wording of this, but the CLI uses its own errors and not Ignition and I didn't want to change this within this PR), which than has the `message` property that includes a better error message which we then can log for the user.

e. g.

```
A ConfigError occured.

Error detected in the production configuration.

Message: The config.production.json file is missing or not valid JSON
Context: Error: /Users/aileennowak/Code/cli-installs/cli_3/config.production.json: ENOENT: no such file or directory, open '/Users/aileennowak/Code/cli-installs/cli_3/config.production.json'

Debug Information:
    Node Version: v6.11.0
    Ghost-CLI Version: 1.0.0-beta.3
    Environment: production
    Command: 'ghost start'
```

### DIFF
--- a/lib/commands/doctor/checks/startup.js
+++ b/lib/commands/doctor/checks/startup.js
@@ -13,10 +13,16 @@ module.exports = [{
     task: (ctx) => {
         let config = Config.exists(path.join(process.cwd(), `config.${ctx.environment}.json`));
 
-        if (config === false) {
+        /**
+        * CASE: `Config.exists()` returns an `options.message` property with a detailed error message
+        * in case an error (e. g. file doesn't exist or has syntaxerrors) occurred.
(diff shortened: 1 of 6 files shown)


acburdine:
Going to close this for now - as per the task in #216 the solution should instead try to point to the specific place in the config where the JSON is invalid.

PR #278, opened 2017-07-05
Add mail config message after installation

cobbspur:
refs #216
- adds a little ui message to guide people towards docs for mail config

### DIFF
--- a/lib/commands/start.js
+++ b/lib/commands/start.js
@@ -61,6 +61,11 @@ class StartCommand extends Command {
                 return this.ui.run(processInstance.enable(), 'Enabling instance startup on server boot');
             }).then(() => {
                 this.ui.log(`You can access your blog at ${instance.config.get('url')}`, 'cyan');
+
+                if (instance.config.get('mail.transport') === 'Direct') {
+                    this.ui.log('\nGhost uses direct mail by default', 'green');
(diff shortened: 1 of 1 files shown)


acburdine:
So one of the side effects of this change is it will be output _every_ time Ghost start is run, regardless of whether or not the user is using direct mail.

IMO this message should be moved into the `config` command, after this line: https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/config/index.js#L64 . There should also probably be a check to see if the user actually _has_ defined any mail, and if so, skip outputting the UI message.

lib/commands/start.js
```diff
@@ -61,6 +61,8 @@ class StartCommand extends Command {
                 return this.ui.run(processInstance.enable(), 'Enabling instance startup on server boot');
             }).then(() => {
                 this.ui.log(`You can access your blog at ${instance.config.get('url')}`, 'cyan');
+                this.ui.log(`\nGhost uses direct mail by default`, `green`);
```
acburdine: Just as a code style note these don't need to be backticks, should probably be normal single quotes just to stay consistent.

cobbspur: haha I was copying the backticks for consistency. lone above has them but can change them happily

cobbspur: roger that, figured that was the case after you mentioned it 👍

PR #285, opened 2017-07-05
fix(ssl): ensure ssl setup works using the --sslemail flag

### LINKED ISSUE #283: SSL doesn't work when using --sslemail
Passing --sslemail to `ghost setup` results in output like this:

```
"[20:11:43] Setting up ssl [started]", 
"[20:11:43] Checking DNS resolution [started]", 
"[20:11:43] Checking DNS resolution [completed]", 
"[20:11:43] Preparing nginx for SSL configuration [started]", 
"[20:11:43] Preparing nginx for SSL configuration [completed]", 
"[20:11:43] Restarting Nginx [started]", 
"Running sudo command: service nginx restart", 
"[20:11:43] Restarting Nginx [completed]", 
"[20:11:43] Getting SSL Certificate [started]", 
"[20:11:50] Getting SSL Certificate [failed]", 
"[20:11:50] \u2192 Command failed: 
```

with 

`"Error: The CA was unable to validate the file you provisioned. ", `

because the `.well-known` location block is missing from the nginx config

acburdine:
closes #283
- cleanup promise handling in ssl setup

- [ ] test on ubuntu

### DIFF
--- a/extensions/nginx/index.js
+++ b/extensions/nginx/index.js
@@ -132,17 +132,23 @@ class NginxExtension extends cli.Extension {
         }, {
             title: 'Preparing nginx for SSL configuration',
             task: (ctx) => {
-                return argv.sslemail ? Promise.resolve({email: argv.sslemail}) : this.ui.prompt({
-                    name: 'email',
-                    type: 'input',
-                    message: 'Enter your email (used for SSL certificate generation)',
-                    validate: value => Boolean(value) || 'You must supply an email'
-                }).then((answer) => {
-                    argv.sslemail = answer.email;
+                let promise;
 
-                    ctx.ssl = {};
+                if (argv.sslemail) {
+                    promise = Promise.resolve(argv.sslemail);
+                } else {
+                    promise = this.ui.prompt({
+                        name: 'email',
+                        type: 'input',
+                        message: 'Enter your email (used for SSL certificate generation)',
+                        validate: value => Boolean(value) || 'You must supply an email'
+                    }).then(answer => { argv.sslemail = answer.email; });
+                }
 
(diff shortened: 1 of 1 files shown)


extensions/nginx/index.js
```diff
@@ -132,23 +132,24 @@ class NginxExtension extends cli.Extension {
...
             task: (ctx) => {
-                return argv.sslemail ? Promise.resolve({email: argv.sslemail}) : this.ui.prompt({
+                let promise = argv.sslemail ? Promise.resolve(argv.sslemail) : this.ui.prompt({
                     name: 'email',
                     type: 'input',
                     message: 'Enter your email (used for SSL certificate generation)',
                     validate: value => Boolean(value) || 'You must supply an email'
-                }).then((answer) => {
-                    argv.sslemail = answer.email;
+                }).then(answer => { argv.sslemail = answer.email; });
```
ErisDS: It took me many reads of this to understand that the problem is the then is attached to or clause.

I think it might be worth expanding this out?

extensions/nginx/index.js
```diff
@@ -132,23 +132,24 @@ class NginxExtension extends cli.Extension {
...
 
-                    ctx.ssl = {};
+                return Promise.all([
+                    promise,
+                    Promise.fromCallback((cb) => NginxConfFile.create(nginxConfPath, cb)).then((conf) => {
```
ErisDS: I have a feeling that using Promise.all is going to lead to more problems?

I'm not 100% sure, but as promise.all doesn't guarantee any order, I guess this code could result in the ssl config having the well-known entry, but not having an email.

If there's an error from the prompt & the config got updated already... what happens if you try to run this again?

acburdine: that's true - was trying to be too efficient 😛

Issue #311, opened 2017-07-07
Add current user to ghost group


sebgie:
To allow the user that has installed ghost to edit files within the `/content` folder it is necessary to add the currently active user to the `ghost` group. This behaviour was removed in #296 because the user needs to log out and in again to activate the new group.

Changes required:
- add user to `ghost` group (removed here: https://github.com/TryGhost/Ghost-CLI/pull/296/commits/c4994806d0d6a37d1ac8634a5ba193d3721d3a19#diff-39682f6dabedcf585eccc2dc12de55d8L42)
- change permissions for the `/content` folder (see #294)
   - #294 is important so that we are not accidentially give execution permissions to an uploaded file. This could lead to an exploit being uploaded via the file upload and executed via a vulnerability in another service running on the same machine.
- add a warning that the user needs to log out/in to acquire group permissions

acburdine:
Closing this as it's replaced by #317

Issue #317, opened 2017-07-08
`ghost exec`/`ghost repair-permissions` commands


acburdine:
Supersedes #311 , related to #307 

In order to facilitate management of an installed Ghost instance with as little pain as possible, while still not messing up file/folder permissions, two new commands needed to be added to the CLI via the `linux` extension:

- [ ] `ghost exec <cmd>`

`ghost exec <cmd>` will essentially work similar to `sudo <cmd>` but after the `sudo` command is run, all permissions will be repaired so that Ghost itself isn't affected.

- [ ] `ghost repair-permissions`

`ghost repair-permissions` will go through the files in the ghost install directory, ensuring that all file permissions are correct

Later:

- [ ] add permissions checks into a `ghost doctor` check

acburdine:
Related: `ghost run` runs Ghost in the foreground, and is useful for debugging why a Ghost instance is not starting. Because of the permissions changes around the linux user spec - we also need to add some modifications to the `ghost run` command.

Reasoning:
- ghost run runs in the foreground, meaning any files created during the process are going to be run as the user that actually runs the process. Example, I have an `acburdine` user and I run `ghost run` as that user, any files created will be owned by `acburdine`. This creates problems for the linux user extension for two reasons:
  - the `content/logs` folder will by owned by `ghost`, so any attempt to write to that directory as another user (e.g. `acburdine`) will fail with write permissions errors
  - if the user runs `sudo ghost run`, it will not run into the same issue, but then any log files will be owned by `root`, meaning if you tried to run `ghost` normally again (e.g. with `ghost start` and systemd), you would get other permissions errors because the `ghost` linux user wouldn't be able to write to the log file. This is the reason why the error in #303 occurred.

Solution:
- if the user is running _on linux_ and the CLI _has configured_ the ghost system user for the instance (e.g. switched the ownership of the `content` folder to the ghost user), then we _always_ need to run `ghost run` _as the ghost user_ so that we don't mess up file permissions.
- To accomplish this, the linux extension will first check if it's running on linux and the CLI has configured the ghost system user.
  - if either are false, the linux user won't do anything, meaning you can run `ghost run` normally without sudo
  - If both are true, and the user is _not_ root (e.g. sudo), then it will error, because otherwise the CLI cannot correctly switch the user id of the process to be that of the ghost user.
[...]

Issue #303, opened 2017-07-06
Ghost log file permissions


kirrg001:
```
> kate@dev-kate-01:~/ghost-kate$ ghost ls
┌────────────────┬──────────────┬──────────────┬──────────────────────┬──────────────────────────────────────┬──────┬─────────────────┐
│ Name           │ Location     │ Version      │ Status               │ Url                                  │ Port │ Process Manager │
├────────────────┼──────────────┼──────────────┼──────────────────────┼──────────────────────────────────────┼──────┼─────────────────┤
│ 207-154-219-31 │ ~/ghost-kate │ 1.0.0-beta.2 │ running (production) │ https://kate-ghost-test.acburdine.me │ 2368 │ systemd
```

```
> kate@dev-kate-01:~/ghost-kate$ ghost log
An error occurred.
Message: 'file does not exist'
```

```
> kate@dev-kate-01:~/ghost-kate$ ghost run
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: EACCES: permission denied, open '/home/kate/ghost-kate/content/logs/https___kate_ghost_test_acburdine_me_production.error.log'
    at Error (native)
```

I had no problem with Ghost log files when i installed/started Ghost with my root user.
`kate` is part of the sudoers group. The whole installation (nginx, ssl, mysql, systemd) worked well.

```kate@dev-kate-01:~/ghost-kate$ ll content/logs/
total 8
drwxrwxr-x 2 ghost ghost 4096 Jul  6 19:13 ./
drwxrwxr-x 7 ghost ghost 4096 Jul  6 19:13 ../
```

acburdine:
The actual permissions error is because you tried to run `ghost run`, which won't work after the permissions have been changed by the ghost user.

However, the first issue of `ghost log` erroring when a log file doesn't exist, is a different one. I can fix that 😄

kirrg001:
I am still hitting this permission error.
Even with your fix for `ghost log`.

```
> ghost start (or ghost run)

events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: EACCES: permission denied, open '/var/www/ghost/content/logs/https___kate_ghost_2_acburdine_me_production.error.log'
    at Error (native)
```

Permissions:
`drwxrwxr-x 2 ghost ghost`

**Using `sudo ghost run` or `sudo ghost start` resolves this.**
But i thought that is not needed? 😕

kirrg001:
[...]

PR #327, opened 2017-07-10
fix(run): better handling of ghost linux user during ghost run

### LINKED ISSUE #307: Content folder configuration
I spent way too much time yesterday to find out where usage of the special content folder is configured. It appears that the content path is configured using an environment variable (https://github.com/TryGhost/Ghost-CLI/blob/22c804cf049dd7319a6bf5d525869afe1768c58c/lib/commands/run.js#L17).

Would it be sensible to put the content path in the configuration file so it is a) obvious and b) harder to get accidentially overwritten?

acburdine:
refs #311
- add handling to linux extension that ensures the user is correct

### DIFF
--- a/extensions/linux/index.js
+++ b/extensions/linux/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const fs = require('fs');
 const os = require('os');
 const execa = require('execa');
 const path = require('path');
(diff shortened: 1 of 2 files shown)


acburdine:
also cc @ErisDS for the wording of the error messages

acburdine:
tested and verified that this works 👍

ErisDS:
Meep. I have no context. What's going on here?

> 'Because you have set up a "ghost" system user, you must run `ghost run` via sudo.'

ErisDS:
^ makes no sense to me because ghost cli sets up a "ghost" system user.

acburdine:
Sorry, more context. Because ghost-cli sets up the "ghost" system user, running `ghost run` won't work due to permissions of the `content` folder. This fixes that problem by setting the group and user of the process to that of the ghost user, but it requires `ghost run` to be initially started using `sudo`, hence the error message.

Updating it to be more clear.

acburdine:
@ErisDS wording is a bit better now (I think)

ErisDS:
> it requires ghost run to be initially started using sudo, hence the error message.

Does "initially" mean "the first time it is run"? 

I'm also confused as to why this closes #307

@sebgie Please intervene here. I don't understand how what's happening in this PR relates to your permissions guide:

> Please be careful here. ghost install is executed without sudo and whenever ghost-cli needs sudo privileges (e.g.: restart Nginx) it will ask for it. Using ghost-cli with sudo could lead to an inconsistent state since all operations would be executed in the context of the super user.

It is so hard right now to keep track of the context of why a decision was made. IMO this issue should be referencing a spec.

acburdine:
It closes #307 because this is kind of a two-fix PR - the first part (the one that closes #307) I will split out into a separate PR to not be as confusing.

First part = saving the content path to the config file instead of mutating process.env
Second part = `sudo ghost run`

This is related to both #303 and #311 - an actual write up of the spec for this change is in [this comment](https://github.com/TryGhost/Ghost-CLI/issues/317#issuecomment-314083764)

acburdine:
The fix for #307 has been removed from this PR and split out into https://github.com/TryGhost/Ghost-CLI/pull/329

sebgie:
I agree with Hannah, given that all ghost-cli commands are run without `sudo` and ask for super user privileges when needed it feels very wrong to suddenly require sudo.

From Slack: https://ghost.slack.com/archives/C1QUCB9T9/p1499681674116302
```
sebastian [12:14] 
just for my understanding, why doesn’t ghost-cli ask for sudo?

austin [12:14] 
because it can't in-process

[12:15] 
it uses `process.setuid` and `process.setgid`

[12:15] 
which you can't use unless you're running as sudo originally
```

Is there a way to let the CLI ask for sudo if needed? I see that run.js uses `child.spawn()` and we could use `sudo -u ghost node current/index.js` in this call to use the `ghost` user?

acburdine:
We _could_, but that would require hardcoding of the Linux extension's behavior into Ghost-CLI itself. I'm willing to do that, but I think that implication does need to be considered.

sebgie:
We are going to make the ghost system user mandatory for Linux systems (#334). To resolve the `ghost run` discussion we are going to have a special runner for Linux.

acburdine:
Closing this as per #334

PR #329, opened 2017-07-10
fix(run): set content path in config rather than in environment

### LINKED ISSUE #307: Content folder configuration

acburdine:
closes #307
- instead of mutating process.env, set the contentPath in config if it does not exist

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -14,7 +14,9 @@ class RunCommand extends Command {
 
         let instance = this.system.getInstance();
 
-        process.env.paths__contentPath = path.join(process.cwd(), 'content');
+        if (!instance.config.has('paths.contentPath')) {
+            instance.config.set('paths.contentPath', path.join(instance.dir, 'content')).save();
+        }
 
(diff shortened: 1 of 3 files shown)


sebgie:
The same environment setting exists in https://github.com/TryGhost/Ghost-CLI/blob/10647318c10738a351f10939b0b87e80635d7f50/lib/tasks/migrate.js#L8.

acburdine:
@sebgie The original reason the env mutating was in `ghost run` itself was because of migrations. Since migrations are run in the same process context, (e.g. `knex-migrator` is run programmatically), it makes more sense I think to mutate the env.

sebgie:
Did that break the tests now? :-/

acburdine:
lol oops - will fix

PR #336, opened 2017-07-10
Pull Linux Extension into CLI proper

### LINKED ISSUE #334: Make system user mandatory for Linux installs
There are no alternatives to creating a system user for Linux installations and therefore this step has to be mandatory.

From Slack (https://ghost.slack.com/archives/C1QUCB9T9/p1499696528020328):
```
hannah [16:22] 
I think the linux user thing doesn't fit well as an extension because it's non-optional

[16:23] 
mysql = optional, can use sqlite3
nginx  = optional, can use apache although not supported by the cli
systemd = optional, can use local process manager, and maybe one day even a custom one
linux user = required otherwise things break
```

acburdine:
refactor: pull linux extension code into CLI proper
closes #334 
- remove linux extension, make it part of the CLI itself

TODOs:
- [x] re-add run functionality from #327 
- [x] test on ubuntu


lib/commands/run.js
```diff
@@ -18,8 +21,70 @@ class RunCommand extends Command {
...
+        let stats = fs.lstatSync(path.join(instance.dir, 'content'));
+
+        if (stats.uid !== ghostuid && stats.gid !== ghostgid) {
+            // folder isn't owned by ghost user, use direct
+            return this.useDirect(instance);
+        }
+
+        let currentuid = process.getuid();
+
+        if (currentuid === ghostuid) {
```
sebgie: Maybe worth adding a comment that this is crucial for systemd?

acburdine: 👍

Issue #313, opened 2017-07-07
[WIP] Ghost-cli output polish


ErisDS:
Today, I started reviewing the default flow of ghost-cli, and have a number of minor, and a few major tweaks I'd like to make to get this ship-shape and ready for the masses 🚢 👨‍👩‍👧‍👦

Below is a screenshot of the current output when running `ghost install` & taking the path of least resistance / official recommendations.

![](https://slack-imgs.com/?c=1&url=https%3A%2F%2Fpuu.sh%2FwDUUM%2F6bdf36e846.png)

This currently includes:
-  17 prompts (?)
-  18 task info messages (ticks) 
-  10 sudo notices (grey)

The sudo notices are really awesome btw ✨ 

IMO - it's fine to have detailed output and the number of task info / sudo notices is not a problem. However the number of user interactions provided by default needs to be reduced and we also need to give users enough context for where they can make a choice, so that they know what to choose!

In addition, there are a few messages which need a little bit of love / tweaking to be clearer or more detailed, most of these are only seen when using --verbose mode.

Please note that in an ideal world, messages would also be output in a different order, see https://github.com/TryGhost/Ghost-CLI/issues/312.

The major points of concern are the options which aren't really optional.

E.g. the 3 prompts which state "Ghost-CLI would like to setup an X file"

At this point, the user has already said yes to setting up the service, IMO the prompt should read 

"Would you like to view the X file?" with the options being (N)o, (y)es, and (a)bort.

I am not 100% sure if these prompts should be present at all, part of me thinks they might be better only appearing in verbose mode? And maybe in verbose mode we show the file by default?! Not sure. The sudo command already outputs the location.

- [x] Improve file prompts

---

There are 2 prompts which can be removed:
- [ ] Do you wish to setup automatic ssl renewal?

If SSL is opted for, there should be no additional prompt for SSL renewal. A message to say this was done is enough.

- [x] Do you wish to enable the Ghost instance to start on reboot?

If systemd is selected, and ghost is started, systemd can be enabled with a message to say its been done, rather than a prompt.

>  Do you wish to set up a ghost system user? Yes

As this is required for systemd, I think the prompt needs removing or reworking, e.g.

- [x] ? Ghost-CLI will create (or use, if it exists) a "ghost" system user, this is required for systemd and recommended for security. Press N to skip system user and systemd setup:

---

- [x] ~~In a bucking of the trend, there is one prompt I'd like to add!!!~~

[...]

PR #337, opened 2017-07-10
🎨 fix(ui): fixes to --start --enable and --quiet flags

### LINKED ISSUE #335: --quiet results in enable/disable being skipped
The start and stop commands both have a `--quiet` flag.

This flag prevents the UI from outputting anything on start/stop, but also prevents the enable/disable step from happening.

I am not sure what the intention of the `--quiet` flag is. It only appears in a handful of places, was it a precursor to `--no-prompt`? 

@acburdine if you can clarify the intention here, I think I can fix/remove this. It's totally non-urgent except in that I am going to be removing the quiet flag when passed to `stop` from `uninstall` so that the systemd disable command is run on `uninstall`.

ErisDS:
refs #313

This removes the prompt for enabling as described in #313 and also deals with the slightly weird start command, as mentioned in https://github.com/TryGhost/Ghost-CLI/pull/246

- the `ghost setup --start` flag
  - The default value for "start" is "prompt"
  - The fallback default value for "start", if it's not allowed to prompt is true
  - Passing --no-start will set start to false, and not prompt
  - Passing --start will set start to true, and not prompt

- the `ghost start --enable` flag
  - Enabling doesn't trigger any prompt
  - The default value for "enable" is true
  - Passing --enable doesn't change anything
  - Passing --no-enable prevents enabling

- calling `ghost uninstall` will now call `ghost stop --disable`

----

This caused a bug with `stopping ghost` outputting multiple times

closes #335

- implement --quiet inside of ui.run

Note: even with this fix, `Starting Ghost` can output multiple times, I think because the promise for processInstance.enabled() can resolve early, and trip up the spinner logic? Not sure....

### DIFF
--- a/lib/commands/setup.js
+++ b/lib/commands/setup.js
@@ -147,7 +147,15 @@ class SetupCommand extends Command {
             }
 
             return this.ui.listr(tasks, {setup: true}).then(() => {
-                if (!argv.prompt || argv.start) {
+                // If we are not allowed to prompt, set the default value, which should be true
+                if (!argv.prompt) {
+                    argv.start = true;
+                }
+
+                // If argv.start has a value, this means either --start or --no-start were explicitly provided
+                // (or --no-prompt was provided, and we already defaulted to true)
+                // In this case, we don't prompt, we use the value of argv.start
+                if (argv.start || argv.start === false) {
                     return Promise.resolve({yes: argv.start});
                 }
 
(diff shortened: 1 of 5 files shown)


lib/commands/setup.js
```diff
@@ -147,7 +147,15 @@ class SetupCommand extends Command {
...
-                if (!argv.prompt || argv.start) {
+                // If we are not allowed to prompt, set the default value, which should be true
+                if (!argv.prompt) {
+                    argv.start = true;
+                }
+
+                // If argv.start has a value, this means either --start or --no-start were explicitly provided
+                // (or --no-prompt was provided, and we already defaulted to true)
+                // In this case, we don't prompt, we use the value of argv.start
+                if (argv.start) {
```
acburdine: if `--no-start` is explicitly provided, `argv.start` will still be false, and this if statement wouldn't execute.

acburdine:
Ah I see why travis is failing on push - the branch got removed 😛

lib/commands/setup.js
ErisDS: Good point, I thought I'd tested all the cases but I guess I missed that one!

PR #338, opened 2017-07-10
fix(ui): 🎨 template prompt adjustments

ErisDS:
Note: I'm not sure the ui.log thing here is the correct pattern. It outputs without a checkmark/tick, and looks a bit weird.

refs #313

- don't allow users to skip, as this would break things
- only show the option to continue, view or edit if --verbose is passed and --no-prompt is not
- reword the question to be would you like to view or edit
- fix the duplicate word "file" for systemd
- add a ui.log for creating the files

- TODO: use a checkmark?!


acburdine:
I'll add a UI optimization that makes checkmarks work

ErisDS:
Fixed up the tests... oh ma gahd there were tests!

PR #341, opened 2017-07-11
Various fixes

acburdine:
These are just a few tiny issues I noticed while testing #338 

- don't prompt for mysql if we shouldn't prompt for mysql (boolean logic is fun)
- make start/no-start work as intended
- log successful template generation regardless if linked to system dir or not.

### DIFF
--- a/lib/commands/config/index.js
+++ b/lib/commands/config/index.js
@@ -80,7 +80,7 @@ class ConfigCommand extends Command {
             });
         }
 
-        if (!argv.db || argv.db === 'sqlite3') {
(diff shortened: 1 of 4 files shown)


acburdine:
@ErisDS my comment on #338 was actually wrong, turns out one of these fixes _did_ have to do with that PR 😛 . The last commit I made moved the log statement up out of the if block, reasoning behind that is `dir` won't be specified if the config file isn't being linked anywhere, but in verbose mode the user will still be asked whether or not they want to create the file....perfectly willing to remove the last commit but the behavior seemed inconsistent.

PR #369, opened 2017-07-20
[WIP] 🐛  fix local installation

### LINKED ISSUE #368: Local dev server does not shut down on correctly, throws fatal error on attempted restart.
### This issue is a

* [X] Bug Report
* [ ] Feature Request

### Summary:

`ghost restart` and `ghost stop`  fail to actually stop the server on a macOS local installation.  This causes a fatal error when attempting to restart Ghost after the first run, due to the port being in use by the original process.  The only way that I've found to shut down the server at this point is to manually find and stop the node process.

### Steps to Reproduce:

1. Set up a local environment using `ghost install local`.
2. Visit localhost:2368 to verify that ghost is running.
3. Run `ghost restart`.  You should see an error that the port is in use.
4. Run `ghost stop`.  You should see a message that no ghost instance is currently running.
5. Run `ghost start`.  You should see an error that the port is in use.
5. Visit localhost:2368 to verify that ghost is *still* running.

### Technical details:

OS: macOS 10.12.5

Debug Information:
    Node Version: v6.11.1
    Ghost-CLI Version: 1.0.0-rc.3
    Environment: development
    Command: 'ghost start'

kirrg001:
closes #368

- if you run `ghost start`, the CLI will spawn a detached process
- this detached process will spawn another process
- the `cp.pid` was wrong in any case
- forward the pid from the spawned process back to parent process

- [x] test on mac
- [ ] test on windows
- [ ] test on linux

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -65,7 +65,7 @@ class RunCommand extends Command {
 
         this.child.on('message', (message) => {
             if (message.started) {
-                instance.process.success();
+                instance.process.success({pid: this.child.pid});
                 return;
             }
 
--- a/lib/utils/local-process.js
+++ b/lib/utils/local-process.js
@@ -2,6 +2,7 @@
 const fs = require('fs-extra');
 const path = require('path');
 const fkill = require('fkill');
+const extend = require('lodash/extend');
 const spawn = require('child_process').spawn;
 const assign = require('lodash/assign');
 const isRunning = require('is-running');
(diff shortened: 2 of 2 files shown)


acburdine:
Closing because this isn't the right solution - and won't work on windows because it relies on IPC messaging to do things correctly, which doesn't work in Windows

acburdine:
Well, as it turns out windows _does_ support ipc messaging - meaning the changes in the previous pr (#366) were wrong 😅 .

However, this is still not the right approach.

acburdine:
sorry wrong button :p

PR #373, opened 2017-07-21
More text-based tweaks 😁

ErisDS:
Raising this just to flag up that I have a handful of very minor text-based tweaks I'd like to make.
These are mostly dependent on docs urls, so need to finish updating docs a bit first.

In this PR so far I have swapped around the checks for pre 1.0.0 and dev installs - I haven't tested this, but short term it seems like pre 1.0.0 is more likely, and so should be the first message.

I'm also adding a link to the migration guide

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -134,6 +134,7 @@ module.exports = {
         // Yargs magic
         // TODO: evaluate if `wrap` is actually needed?
         yargs.wrap(Math.min(100, yargs.terminalWidth()))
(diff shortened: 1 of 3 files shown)


acburdine:
👍  Looks good so far

ErisDS:
Think this one is done

ErisDS:
NVM failed tests... looking

acburdine:
Tests are failing because `yargs.epilogue` needs to be stubbed in the unit tests for the command class

ErisDS:
Yar fixed 😁

acburdine:
*yargs fixed :trollface:

PR #374, opened 2017-07-21
fix(help): explicit setup stages

ErisDS:
@acburdine not sure if this is controversial?


acburdine:
Unless I'm mistaken this will only allow running one setup stage at a time?

ErisDS:
Yeah I wasn't sure if this was purely "output" or if it actually changed behaviour. 

If it changes behaviour then I have a better option... 2 secs

acburdine:
Closing in favor of #376

PR #376, opened 2017-07-21
fix(help): explicit setup stages

ErisDS:
V2 - maybe less controversial 😬


acburdine:
I get why this helps a lot, but I'm still not _super_ excited about it because it hardcodes extension behavior into the CLI itself.

Probably fine for now though - whenever a better method of determining setup stages can be found/implemented then this can be replaced with whatever that ends up being

PR #380, opened 2017-07-25
feat(help): status alias

jloh:
After seeing #375 I realised I find myself constantly having to lookup the `ls` command when I really just want to see the `status` of the running instances.

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -16,6 +16,7 @@ const extraAliases = {
     busters: 'buster',
     intsall: 'install',
     insall: 'install',
+    status: 'ls',
     upgrade: 'update',
     udpate: 'update'
 };


lib/commands/ls.js
```diff
@@ -23,5 +23,6 @@ class LsCommand extends Command {
 
 LsCommand.description = 'View running ghost processes';
 LsCommand.global = true;
+LsCommand.aliases = ['status'];
```
acburdine: Based on how Ghost-CLI commands are loaded/invoked, adding aliases here won't work unfortunately.
The alias should instead be added here: https://github.com/TryGhost/Ghost-CLI/blob/master/lib/bootstrap.js#L13 - it's not the greatest solution for aliases but it's the only one that keeps the CLI from having to require all of the command files before running the one it needs.

acburdine: After seeing #375 I see how it could be _really_ confusing - I didn't realize the implications of aliases in individual command files until _after_ that PR was merged 😬

jloh: Haha I just followed that MR and didn't see the follow on commit!

Will change this over to the new syntax now. Vim is highlighting `status` as a var or command or some sort, I assume I have to quote it?

jloh:
Changes made!

lib/commands/ls.js
acburdine: No, status should be a valid object key w/o quotes, not sure why Vim would be highlighting it.

lib/bootstrap.js
```diff
@@ -16,6 +16,7 @@ const extraAliases = {
     busters: 'buster',
     intsall: 'install',
     insall: 'install',
+    'status': 'ls',
```
acburdine: Status doesn't need to be quoted - this will fail linting because of that 😄

jloh: Didn't see your other comment before pushing those changes up, now unquoted in all its glory.

ErisDS:
> After seeing #375 I see how it could be really confusing - I didn't realize the implications of aliases in individual command files until after that PR was merged 

@acburdine https://github.com/TryGhost/Ghost-CLI/commit/fa3e6c8e5c2c813708103dd23ef3ae0bcc1fefec also doesn't explain - what are the implications?!

acburdine:
Sorry implications was the wrong choice of wording - I didn't realize when I merged it that #375 wouldn't work based on how the command stack is invoked.

If commands defined their own aliases inside each command class, then each command file would have to be manually required in order to get the list of aliases before running the one single command that needs to run. This has a big hit on performance, esp. for the more complicated commands that require other utils/packages/etc. Ghost-CLI gets around that by knowing about all the aliases beforehand, and only requiring the single command file that is needed.

jloh:
> what are the implications?!

https://www.youtube.com/watch?v=MZ1lc6KASWg

PR #401, opened 2017-07-29
fix(config): transform urls to lowercase

### LINKED ISSUE #399: Always store URLs in lowercase
Refs https://github.com/TryGhost/Ghost/issues/8772.

Urls are lowercase. They can be uppercase, but that will likely confuse a lot of systems, not just Ghost Admin in places. 

If a user enters their url with uppercase characters, e.g. www.MyAwesomeWebsite.com ghost cli should store the value in lowercase.

This is somewhere halfway between a bug, a feature request, and so trivial it doesn't really matter 😂

ErisDS:
@austin - I couldn't get `coerce` to work, no matter _where_ I put it! Dunno if you can find a more elegant solution for this?

closes #399

- I tried to do this with the yargs coerce method but couldn't find anywhere that it worked
- Instead, have implemented a transform method and added toLowerCase for urls

### DIFF
--- a/lib/commands/config/advanced.js
+++ b/lib/commands/config/advanced.js
@@ -24,6 +24,7 @@ module.exports = {
         description: 'Blog URL with protocol E.g. http://loveghost.com',
         validate: value => validator.isURL(value, {require_protocol: true}) ||
             'Invalid URL. Your URL should include a protocol, E.g. http://my-ghost-blog.com',
+        transform: value => value.toLowerCase(),
         type: 'string',
         group: 'Ghost Options:'
     },
@@ -32,6 +33,7 @@ module.exports = {
         configPath: 'admin.url',
         validate: value => validator.isURL(value, {require_protocol: true}) ||
             'Invalid URL. Your URL should include a protocol, E.g. http://my-ghost-blog.com',
+        transform: value => value.toLowerCase(),
         type: 'string',
         group: 'Ghost Options:'
     },
--- a/lib/commands/config/index.js
+++ b/lib/commands/config/index.js
@@ -35,6 +35,10 @@ class ConfigCommand extends Command {
                 });
             }
 
+            if (value && option.transform) {
+                value = option.transform(value);
+            }
+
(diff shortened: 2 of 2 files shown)


lib/commands/config/advanced.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
         description: 'Blog URL with protocol E.g. http://loveghost.com',
         validate: value => validator.isURL(value, {require_protocol: true}) ||
             'Invalid URL. Your URL should include a protocol, E.g. http://my-ghost-blog.com',
+        transform: value => { return value.toLowerCase(); },
```
acburdine: gonna be a nitpick - can this be `value => value.toLowerCase();` in keeping with the shorthand syntax above?

lib/commands/config/index.js
```diff
@@ -46,7 +46,7 @@ class ConfigCommand extends Command {
                     }));
                 }
 
-                this.instance.config.set(configKey, value);
+                this.instance.config.set(configKey, option.transform ? option.transform(value) : value);
```
acburdine: I wonder if transform shouldn't be called before validate? I'm open to either place but I feel like transforming before validation would be a bit better? That way transform doesn't leave us with an invalid value for whatever reason.

lib/commands/config/advanced.js
ErisDS: 👍  done

lib/commands/config/index.js
ErisDS: Good point. It's always the easiest ones haha

ErisDS: Changed how it works now

lib/commands/config/index.js
```diff
@@ -35,6 +35,10 @@ class ConfigCommand extends Command {
                 });
             }
 
+            if (value && options.transform) {
+                value = options.transform(value);
```
acburdine: needs to be option.transform :wink:

ErisDS:
This is why midnight bugfixes are a bad idea.

Issue #281, opened 2017-07-05
Ghost-CLI: npm install sqlite3 --save error


johnathon-b:
Currently when using ghost-cli 1.0.0.4 and using `ghost install local` during the installation I get this  error and the installation fails. 

![image](https://user-images.githubusercontent.com/18176012/27876943-7f5c6e42-617e-11e7-9a15-6dea7422927a.png)

johnathon-b:
This was fixed by uninstalling and reinstall of ghost-CLI.

`npm un -g ghost-cli && npm i -g ghost-cli`

ErisDS:
For anyone else who comes across this issue, it is caused by changing node versions after installing ghost-cli. 

Sqlite3 is a binary dependency that comes with precompiled binaries and downloads the right one for your system on install. If you change the environment, it will look for a different binary, which wasn't downloaded.

ltribolet:
I tried to remove everything (purge nodejs from system and remove all node_modules across system), so this is like a fresh install and I've still got the error.

```
$ npm ls -gp --depth=0 | awk -F/ '/node_modules/ && !/\/npm$/ {print $NF}' | xargs npm -g rm
$ sudo apt-get remove --purge nodejs
$ sudo apt-get install nodejs
$ npm install -g ghost-cli
```

Ubuntu 16.04
Node v6.11.0

```
➜  blog ghost install local
✔ Checking for latest Ghost version
✔ Running system checks
✔ Setting up install directory
✔ Downloading and installing Ghost v1.0.0-beta.2
✔ Moving files
✔ Configuring Ghost
ℹ Running setup checks [skipped]
✔ Setting up instance
Knex: run
$ npm install sqlite3 --save
Error: Cannot find module '/usr/lib/node_modules/ghost-cli/node_modules/sqlite3/lib/binding/node-v48-linux-x64/node_sqlite3.node'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/usr/lib/node_modules/ghost-cli/node_modules/sqlite3/lib/sqlite3.js:4:15)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
[...]

PR #408, opened 2017-07-31
Improve sqlite error message

acburdine:
refs #281 

cc @ErisDS for wording

### DIFF
--- a/lib/tasks/migrate.js
+++ b/lib/tasks/migrate.js
@@ -38,7 +38,7 @@ module.exports = function runMigrations(context) {
     return knexMigratorPromise.then(() => {
         config.set('logging.transports', transports).save();
     }).catch((error) => {
-        if (error.stderr.match(/CODE\: ENOTFOUND/)) {
+        if (error.stderr && error.stderr.match(/CODE\: ENOTFOUND/)) {
             // Database not found
             error = new errors.ConfigError({
                 config: {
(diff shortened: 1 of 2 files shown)


lib/tasks/migrate.js
```diff
@@ -56,6 +56,12 @@ module.exports = function runMigrations(context) {
                 message: 'Invalid database username or password',
                 environment: context.instance.system.environment
             });
+        } else if (error.stdout && error.stdout.match(/npm install sqlite3 \-\-save/)) {
+            // We check stdout because knex outputs to stdout on this particular error
+            error = new errors.SystemError(
```
ErisDS: How come SystemError doesn't take an Object, but ConfigError does?

BTW: I'm working on a docs update to link to here 2 mins

acburdine: All errors can either take an object or a string - the string is a shorthand if you just need an error message.

ConfigErrors have additional properties needed to add more help information, which is why an object is passed, whereas SystemErrors are simple enough to just require a string message.

ErisDS: https://docs.ghost.org/v1.0.0/docs/troubleshooting#section-sqlite3-install-failure

If you have a second to write out the `ghost config`  command that would switch a user from sqlite3 to mysql that'd be super handy

PR #409, opened 2017-07-31
fix(doctor): add install checks for various nvm edge cases

### LINKED ISSUE #281: Ghost-CLI: npm install sqlite3 --save error

acburdine:
closes #281
- skip directory check if `--no-setup-linux-user` option is passed
- throw error if npm bin directory is not the same as the one used to install ghost-cli

TODO:
- [ ] test this on ubuntu

### DIFF
--- a/lib/commands/doctor/checks/install.js
+++ b/lib/commands/doctor/checks/install.js
@@ -17,7 +17,7 @@ const eol = os.EOL;
 
 const tasks = {
     // While it's not an actual task, we put it here to make it easier to test
-    checkDirectoryAndAbove: function checkDirectoryAndAbove(dir) {
+    checkDirectoryAndAbove: function checkDirectoryAndAbove(dir, extra) {
         if (isRoot(dir)) {
             return Promise.resolve();
         }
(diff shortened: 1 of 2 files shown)


lib/commands/doctor/checks/install.js
```diff
@@ -13,28 +13,39 @@ const isRoot = require('path-is-root');
...
 
-    return checkDirectoryAndAbove(path.join(dir, '../'));
+        return checkDirectoryAndAbove(path.join(dir, '../'), extra);
+    });
 }
 
 module.exports = [{
     title: 'Checking system Node.js version',
-    task: () => {
+    task: (ctx) => {
```
ErisDS: Do we always need to run this check? It seems to only be important if using sqlite3... or is it also impacting DTrace>Bunyan>Knex Migrator?
DTrace is an optional dep as far as I know, so if this only impacts sqlite3 adding this as a global check feels like it's going to trip up users who are already setup...

In my opinion, it's the kind of thing that we should do _after_ detecting a failure with sqlite3 install so that we can give an exact error message. Is that possible?

acburdine: That's definitely possible - however if a user gets into this state and doesn't trigger the sqlite message, updates to ghost-cli wouldn't take effect.

Example:
- user installs ghost-cli globally
- user installs nvm after that
- user sets up install w/systemd (which makes the "ghost" system user use the globally-installed ghost-cli version
- user installs new ghost-cli version (when will end up in the nvm directory)
- systemd service file still gets old ghost-cli version, with any potential bugs/missing features, etc.

ErisDS:
We can keep this as-is and see how it goes. Defo requires a minor bump tho ;)

acburdine:
Moved part of this into #412 -> the nvm checks will be the sole focus of this pr

acburdine:
cc @ErisDS for a final wording check on this

ErisDS:
Totally fine with me - I will probably improve it later based on feedback + once I've got a decent section in the docs. Still struggling with how to explain this.

PR #411, opened 2017-07-31
fix(doctor): fix prompting and error handling with --no-prompt

### LINKED ISSUE #410: Running `ghost install` with `--no-prompt` option fails if mysql is not installed
See https://github.com/TryGhost/Ghost-CLI/pull/318#issuecomment-319064657

We should throw an error and exit rather than prompt if `--no-prompt` is passed as an option.

acburdine:
closes #410
- throw errors instead if prompts are disabled
- fix error if remote mysql host is used

TODO: tests


ErisDS:
LGTM 👍

PR #420, opened 2017-08-02
Improve Systemd checks

### LINKED ISSUE #414: systemd unit file location
### This issue is a

* [x] Bug Report
* [ ] Feature Request

### Summary

ghost-cli creates systemd unit files in `/lib/systemd/system/`.
According to my understanding the directory `/lib/systemd/system/` is intended for system-supplied unit files, i.e. those installed by the package manager of the distribution.
In contrast to that the directory `/etc/systemd/system/` is intended for local machine-specific unit files (see [here](https://www.freedesktop.org/software/systemd/man/systemd.unit.html)). I feel that the unit files installed by ghost-cli clearly fall in the local and machine-specific category and I was surprised to see them end up in `/lib/systemd/system/`.

This is more of a minor complaint than a proper bug report, but I think it would be good to follow best practices.

### Technical details (will be automatically output by Ghost-CLI if an error occurs):

* OS: Ubuntu 16.04
* Node Version: v6.11.1
* Ghost-CLI Version: v1.0.1

acburdine:
closes #414 

TODO:
- [x] test on ubuntu
- [x] add unit tests

### DIFF
--- a/extensions/systemd/get-uid.js
+++ b/extensions/systemd/get-uid.js
@@ -22,7 +22,7 @@ module.exports = function getUid(dir) {
 
         return uid;
     } catch (e) {
-        if (!e.message.match(/no such user/)) {
+        if (!e.message.match(/no such user/i)) {
             throw e;
         }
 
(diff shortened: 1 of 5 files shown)


acburdine:
Soo, after testing this I discovered that systemctl enable doesn't work anymore - which obviously isn't ideal 😕

acburdine:
Removed the switch to /etc/systemd - this PR now just adds tests and checks file instead of cliConfig property

PR #424, opened 2017-08-03
fix(config): 🐛  don't enforce case for mail config

### LINKED ISSUE #421: Nodemailer "sendmail" transport is case sensitive: change in behavior
https://github.com/nodemailer/nodemailer/tree/0.7#setting-up-sendmail

As you can see, they use an all lowercase "sendmail" in the definition.  I have used this for all my ghost installs until I moved to 1.0.0 release (1.0.3 specifically). Apparently Ghost-CLI now does validation that expects "Sendmail" and rejects "sendmail".  There is nothing on the nodemailer github page that dictates case, and "sendmail" worked with older ghost, other nodemailer instances I have used, and all of the alpha/beta/RC instances of Ghost I ran (pre 1.0 Ghost-CLI).

Looking at https://github.com/nodemailer/nodemailer/blob/0.7/lib/transport.js they use toUpper to compare transport settings anyway.

ErisDS:
TODO:

- [x] test this once we get past the 1.4 issues

closes #421

- uses toLowerCase() and lists all config options in lowercase

### DIFF
--- a/lib/commands/config/advanced.js
+++ b/lib/commands/config/advanced.js
@@ -8,9 +8,9 @@ const url = require('url');
 
 const BASE_PORT = '2368';
 const knownMailServices = [
-    'AOL', 'DynectEmail', 'FastMail', 'Gmail', 'GoDaddy', 'GoDaddyAsia', 'GoDaddyEurope', 'hot.ee', 'Hotmail', 'iCloud',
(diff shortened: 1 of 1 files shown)


ErisDS:
Have tested this now, seems ok

PR #427, opened 2017-08-03
chore(test): add test stubs for remainder of files

acburdine:
- this gives us a more accurate representation of what the repo test coverage actually is 🙈
- include extension tests in test run


acburdine:
well - now we have the true test coverage 🙈

Issue #428, opened 2017-08-03
Print ghost help information faster


joehand:
### This issue is a

* [ ] Bug Report
* [x] Feature Request

### Summary

Running `ghost` or `ghost help` is quite slow to print out the information. It'd be great if it printed out quicker. 

#### Requests

* Improve speed of `ghost help` & `ghost`
* Have `ghost` with no arguments print out the help information.

On my server, `ghost help` takes over 4 seconds to print out information:

```
$ time ghost help

<help info>

real	0m4.183s
user	0m3.668s
sys	0m0.488s
```

I had run `ghost` first, which also took a few seconds to tell me to run `ghost help`.

We had the same issue in https://github.com/datproject/dat and were able to [greatly improve our output time](https://github.com/datproject/dat/issues/820) by moving requires for each subcommand [inside the command functions](https://github.com/datproject/dat/blob/master/src/commands/create.js#L20-L29), so they were only required if that command was run & the function called. You may also find https://github.com/maxogden/require-times to be useful debugging.

I didn't see at a quick glance where the slow requires are happening but happy to try to look a bit more if needed.

Thanks! 💖

acburdine:
Hey @joehand! So here's a couple of things:

1) We already only require individual commands if they're run - ghost help is actually the _only_ command that requires all of them.
2) Depending on the size of your global node_modules folder, command discovery can actually be slower than requiring. Ghost-CLI doesn't necessarily know about all of the commands initially, as there is support for extensions to define their own commands. Discovering those extensions, and then requiring any commands within said extensions, can be slower depending on the filesystem

I can look at moving some of the requires a bit lower into the run functions, however given that you said it took several seconds for `ghost` without any arguments to tell you to run `ghost help`, my guess is that the slowness is coming from the command discovery part rather than the actual requiring of the command files. (Note: if you want to see where the slow parts are, running `DEBUG=ghost-cli:* ghost help` will actually show you which parts are slower 😉 )

joehand:
Sweet! It does seem the server is a bit slow. The command was much faster on my local machine.

It looks like there is a lot of time happening before the first debug statement prints (see total time vs each individual timestamp), so hopefully the lazy loading will help. Here output from the debug on my server:

<details>
```
  ghost-cli:find-extensions searching for extensions +0ms
  ghost-cli:bootstrap Found 3 extensions: ghost-cli-nginx, ghost-cli-mysql, ghost-cli-systemd +107ms
[...]

PR #429, opened 2017-08-04
perf: lazily require modules in commands

acburdine:
refs #428
- defer requiring modules in commands to as late as possible


acburdine:
Coverage decreased because all of the lines at the top of command files technically counted as lines covered since the files were required - now that they're moved farther down they don't count anymore 🙈

PR #454, opened 2017-08-17
Offer better default database name when installing

### LINKED ISSUE #453: ghost install default database name causes installations to share a database instance
### This issue is a

* [x] Bug Report
* [x] Feature Request

### Summary

If you run `ghost install` (to do a production installation) then it prompts you to create a mysql database called `ghost_production`.

If you run another installation, it also suggests the same database name. When I accepted the default twice in a row, I ended up with two blog installations sharing the same database and chaos ensued.

I reckon it would be better to include some kind of identifier in the default database name, so I can (probably) accept the defaults. Given the point in time at which the default prompt values are calculated, it's best to use the local folder name. e.g. if you are installing in `/var/www/tom` then the database name should be `ghost_production_tom`.

I have already fixed this in my branch.

1. run `ghost install` and accept the default database name
2. repeat without changing the database name.

I reckon this is *kind of* a bug. It's certainly kinder if we offer a default name which is more likely to be unique. Making it definitely be unique is quite complicated and is a possible future enhancement.

### Technical details (will be automatically output by Ghost-CLI if an error occurs):

* OS: 
* Node Version:
* Ghost-CLI Version: 
* Environment: 
* Command:

TomChantler:
Names the default mysql database to be `ghost_environment_foldername`.

e.g. If you run `ghost install` in `/var/www/tom` then the default database name will be `ghost_production_tom`.

closes #453

### DIFF
--- a/lib/commands/config/index.js
+++ b/lib/commands/config/index.js
@@ -71,6 +71,8 @@ class ConfigCommand extends Command {
 
     getConfigPrompts(argv) {
         const validator     = require('validator');
+        const path = require('path');
+
         let prompts = [];
 
         if (!argv.url) {
@@ -119,7 +121,7 @@ class ConfigCommand extends Command {
                     type: 'input',
                     name: 'dbname',
                     message: 'Enter your Ghost database name:',
-                    default: this.instance.config.get('database.connection.database', `ghost_${this.system.environment}`)
+                    default: this.instance.config.get('database.connection.database', `ghost_${this.system.environment}_${path.basename(process.cwd())}`)
                 });
             }
         }


lib/commands/config/index.js
```diff
@@ -119,7 +121,7 @@ class ConfigCommand extends Command {
                     type: 'input',
                     name: 'dbname',
                     message: 'Enter your Ghost database name:',
-                    default: this.instance.config.get('database.connection.database', `ghost_${this.system.environment}`)
+                    default: this.instance.config.get('database.connection.database', `ghost_${this.system.environment}_${path.resolve(process.cwd()).substr(path.resolve(process.cwd()).lastIndexOf('/') + 1)}`)
```
acburdine: instead of the `path.resolve` stuff, you can just do `path.basename(process.cwd())` and it will get you the last section of the path 😁

TomChantler: Ah, I didn't know that. Seems more sensible.

vikaspotluri123: Thinking out loud, would it be possible to use the instance name rather than the folder name, like the service?

acburdine: @vikaspotluri123 possibly - although at this point none of the validation has been run on any of the arguments. We'd technically have access to the argv.name command line arg but the default service name (if none is provided) is based off of the URL - which wouldn't necessarily have been defined at this point.

TomChantler: There isn't an instance name at the point at which the prompts are created. That's why I referenced it as a possible future enhancement

PR #465, opened 2017-08-28
Update migration docs URL

vikaspotluri123:
- Fixes TryGhost/Ghost#8948
- Using v0.x URL (https://docs.ghost.org/v0.11/docs/migrating-to-ghost-version-100) over v1.x (https://docs.ghost.org/v1/docs/migrating-to-1-0-0) since the install is assumed to be 0.x

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -32,7 +32,7 @@ function checkValidInstall(name) {
         console.error(
             chalk.yellow('Ghost-CLI only works with Ghost versions >= 1.0.0.'),
             '\nIf you are trying to upgrade Ghost LTS to 1.0.0 ' +
-            `please see ${chalk.blue.underline('https://docs.ghost.org/v1/docs/migrating-to-1-0-0')}.`,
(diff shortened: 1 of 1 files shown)


lib/command.js
```diff
@@ -32,7 +32,7 @@ function checkValidInstall(name) {
         console.error(
             chalk.yellow('Ghost-CLI only works with Ghost versions >= 1.0.0.'),
             '\nIf you are trying to upgrade Ghost LTS to 1.0.0 ' +
-            `please see ${chalk.blue.underline('https://docs.ghost.org/v1/docs/migrating-to-1-0-0')}.`,
+            `please see ${chalk.blue.underline('https://docs.ghost.org/v0.11/docs/migrating-to-ghost-version-100')}.`,
```
acburdine: FWIW the 1.0 documentation has more in it - therefore think using the v1 migration doc is better. What needs to be fixed is the url - which is actually https://docs.ghost.org/v1/docs/migrating-to-ghost-1-0-0.

PR #466, opened 2017-08-28
Update typos in docs URLs

vikaspotluri123:
- Normalized explicit versioning (from `1.0.0` to `1.0`)
- docs and v1 was flipped for another URL

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 
 ## Documentation
 
-- [Complete Setup Guide](https://docs.ghost.org/docs/install)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -19,8 +19,8 @@
 ## Documentation
 
 - [Complete Setup Guide](https://docs.ghost.org/docs/install)
```
vikaspotluri123: Does this need to be normalized as well?

acburdine: probably, yeah

Issue #468, opened 2017-08-30
ghost backup


ChaosMTA:
### This issue is a

* [ ] Bug Report
* [ x] Feature Request

### Summary

Addition of a `ghost backup` command which gathers a dump of the MySQL database as well as a tar of the content folder would be a great addition. Users could add it to a cron task and backup the files to whatever 3rd party they wish.

I discussed this a bit on the slack with Austin and we think it would be a useful addition.

acburdine:
Braindump:
These commands would be useful - might also be worth adding some extension hooks to define _where_ exactly to send the backups. The default implementation in core would be just saving it to the local filesystem, other extensions could potentially save to s3, dropbox, etc.

vikaspotluri123:
@acburdine can you assign me to this please :)

ErisDS:
cc @sebgie & @kirrg001 

I know there's a lot of active work going on here, which I'm super excited about! I wanted to flag up that we had a chat about this yesterday in a wider meeting about backups where we ended up talking about whether Ghost should have a backup mechanism. 

It made me realise that there are a couple of different things that `ghost backup` could possibly do. 

Where we are:

The original spec on this issue is for a mysqldump + a tar of the content folder. This sounds a lot like what we were discussing yesterday. Meanwhile, it seems the implementation is doing an export of the database to JSON and then a zip of the content folder. These small differences are, I think quite important. 

Also in both the spec & the implementation, the job of `ghost backup` is to create the backup file - whatever that is, and then it would be the job of someone adding this to a cron to determine where the backup goes. 

----

Where I'd like to be:

I think it is correct that `ghost backup` should only prepare the backup, and that this should then be usable in configuring a cron to run the command & move the backup file to a 3rd party. However, I think this scenario should be fully tested before shipping the feature, to make sure that it is compatible with being used that way (E.g. with AWS s3 and/or DO block storage) - as this is definitely the most likely use case. I'd also like to end up with how to do this clearly documented. 

In terms of what the backup does, I think it would be wrong to overlook using mysqldump instead of, or as-well-as doing a JSON backup. It may be that this needs to be coded so it's a possible future extension?  

I'm also not 100% on the method currently being used to generate the JSON export. Currently the exporter is being required, which seems odd when we have a backup mechanism in Ghost which could either be required or used over the API. There is some code duplication I'd like to avoid, and it's fine IMO if that needs changes to Ghost itself.

[...]

PR #474, opened 2017-09-12
[HOLD]✨Backup subcommand

vikaspotluri123:
Adds the ability to run `ghost backup` to create a zip archive of your important data - the config file, your content folder, and your database

ref: #468 

# Todo List
- [x] Create functions for each core task
- [x] Add UI logging
- [x] Write test suite
- [x] Integrate w/ extensions


lib/commands/backup.js
```diff
@@ -0,0 +1,109 @@
...
+class BackupCommand extends Command {
+    run(argv) {
+        const instance = this.system.getInstance();
+        const config = instance.config.values;
+        const datetime = (new Date()).toJSON().substring(0, 10);
+
+        let exporter;
+        let saveLocation = process.cwd();
+
+        // @todo check if we need to ensure we're in a ghost directory
```
acburdine: You don't need to check - all commands by default check for a valid install directory unless there's a static "global" property that's set to true. (for example: https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/ls.js#L26 - the line disables the valid install check.)

tl;dr you don't need to add a check 😁

lib/commands/backup.js
```diff
@@ -0,0 +1,109 @@
...
+
+        if (argv.output) { // @todo add o as an alias
```
acburdine: this todo can be removed as well - you've added `o` as an alias down below, and yargs handles populating the args object correctly using aliases, so if you specify `--output <asdf>` or `-o <asdf>`, the args object will have both `o` and `output` populated with the value in both cases.

lib/commands/backup.js
```diff
@@ -0,0 +1,109 @@
...
+            saveLocation = path.isAbsolute(argv.output) ? argv.output : path.join(process.cwd(), argv.output);
+            fs.ensureDirSync(saveLocation);
+        }
+
+        saveLocation = path.join(saveLocation, `ghost.backup.${datetime}.zip`)
+
+        instance.checkEnvironment();
+
+        if (instance.running()) {
```
acburdine: I'd change the logic a bit here:

```js
if (instance.running()) {
  this.ui.log() // what you have for this line is fine
  instance.loadRunningEnvironment(); // this line will load the environment based on which one is running
} else {
  instance.checkEnvironment(); // if Ghost isn't running, do the checkEnvironment call
}
```

lib/commands/backup.js
```diff
@@ -0,0 +1,109 @@
...
+            this.ui.log('Ghost is currently running. Backing up might take longer and slow down your site','yellow');
+        }
+
+        const backupData = {
```
acburdine: so honestly this whole part isn't necessary - if you save the `.ghost-cli` file (which you are doing already), then most of this data is redundant.

lib/commands/backup.js
```diff
@@ -0,0 +1,109 @@
...
+            console.log(e);
+            return Promise.reject(new Error('Unable to initialize database exporter'));
+        }
+
+        // @todo: Log this?
+        debug('Creating database backup');
+
+        const props = {
+            database: exporter.doExport({}),
+            environment: this.system.environment
```
acburdine: not sure why this is in the props - it's not a promise and you've got access to the environment inside the `then` callback regardless

lib/commands/backup.js
```diff
@@ -0,0 +1,109 @@
...
+        };
+
+        return Promise.props(props).then((data) => {
```
acburdine: this can just be `return exporter.doExport({}).then((data) => {`

lib/commands/backup.js
```diff
@@ -0,0 +1,109 @@
...
+                // Add database
+                zipFile.addFile('database.json', Buffer.from(JSON.stringify(data.database)));
+                // Add .ghost-backup
+                zipFile.addFile('.ghost-backup', Buffer.from(JSON.stringify(backupData)));
+
+                // Write zip file to cwd
+                zipFile.writeZip(saveLocation);
+                debug('Wrote zipFile')
+
+                return Promise.resolve(saveLocation);
```
acburdine: not sure why this passed in the resolve

lib/commands/backup.js
```diff
@@ -0,0 +1,139 @@
...
+        }, {
+            title: `Writing backup file to ${this.saveLocation}`,
+            task: this.writeZipFile.bind(this)
+        }]);
+    }
+
+    backupContent() {
+        // Add content folder // @todo: Do we need to check if the content folder exists?
+        // @todo: Figure out how to deal with Windows bug
+        this.zipFile.addLocalFolder(path.join(process.cwd(), 'content/'), 'content/');
```
acburdine: So yeah, the windows bug isn't great. However, we should be able to fairly easily use another lib to recursively traverse the content folder and add it to the zip.

This library looks pretty good: https://github.com/jprichardson/node-klaw

lib/commands/backup.js
```diff
@@ -0,0 +1,178 @@
...
+
+const debug = createDebug('ghost-cli:backup');
+
+function absolutePath(inputPath) {
+    return path.isAbsolute(inputPath) ? inputPath : path.join(process.cwd(), inputPath);
+}
+
+class BackupCommand extends Command {
+    run(argv) {
+        return this.ui.listr([{task: this.initialize.bind(this),
```
acburdine: Because I'm a bit OCD 😁  (and it makes it a bit easier to read) can you split this out into multiple lines, e.g.
```js
return this.ui.listr([{
  task: this.initialize.bind(this),
  title: 'Checking Configuration'
}], {argv: argv}).then(() => {
  // omitted code
});
```

lib/commands/backup.js
```diff
@@ -0,0 +1,178 @@
...
+        return Promise.resolve();
+    }
+
+    backupContent() {
+        let files;
+        try {
+            files = walker('./content', {nodir: true});
+        } catch (E) {
+            debug(`Failed reading content folder: ${E}`);
+            return Promise.reject(new Error('Failed to read content folder'));
```
acburdine: This should probably be a SystemError rather than a regular error, mainly because regular errors will output a stack trace into the log file, and we don't necessarily need one here.

package.json
```diff
@@ -53,6 +53,7 @@
     "ghost-ignition": "2.8.14",
     "inquirer": "3.2.3",
     "is-running": "2.1.0",
+    "klaw-sync": "^3.0.0",
```
acburdine: Can this be a pinned version? e.g. `3.0.0` rather than `^3.0.0`.

acburdine:
This looks really good 😄 I'll pull this down and test later tonight!

ChaosMTA:
Awesome! @vikaspotluri123

vikaspotluri123:
👍 done

acburdine:
@vikaspotluri123 can you go ahead and rebase this? pulling it down and testing today but it says there are merge conflicts :/

lib/commands/backup.js
```diff
@@ -0,0 +1,200 @@
...
+            title: 'Backing up config files',
+            task: this.backupConfig
+        }, {
+            title: 'Backing up content folder',
+            task: this.backupContent
+        }, {
+            title: 'Writing backup file',
+            task: this.writeZipFile
+        }],{
+            env: this.system.environment,
```
acburdine: This causes an issue later on - because you run the `checkEnvironment` function below, the `system.environment` property has the potential to get updated, and therefore the value of `env` would be different.

lib/commands/backup.js
```diff
@@ -0,0 +1,200 @@
...
+            dir = path.relative(cwd, dir);
+
+            try {
+                ctx.zipFile.addFile(`${dir}/${name}`, readFile(`./${dir}/${name}`), '', 644);
+            } catch (err) {
+                ctx.instance.ui.log(`Failed to backup ${dir}/${name}`, 'yellow');
+            }
+        });
+
+        return Promise.resolve();
```
acburdine: You don't _have_ to return a promise here 😄 If a method is synchronous then it doesn't need a promise.

However, given that readfile can potentially take some time depending on the size of the file, my advice actually would be to use the Promise.all method - do something like:

```js
const promises = files.map((file) => {
  const location = file.path.replace(/\\/g, '/');
  let dir = path.dirname(location);
  const name = path.basename(location);
  dir = path.relative(cwd, dir);

  return fs.readFile(`./${dir}/${name}`).then((contents) => {
    ctx.zipFile.addFile(`${dir}/${name}`, contents, '', 644);
  }).catch((error) => {
    ctx.instance.ui.log(`Failed to backup ${dir}/${name}`, 'yellow');
  });
});

return Promise.all(promises);
```

lib/commands/backup.js
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        const cwd = process.cwd();
+        const readFile = fs.readFileSync;
+
+        files.forEach((file) => {
+            const location = file.path.replace(/\\/g,'/');
+            let dir = path.dirname(location);
+            const name = path.basename(location);
+            dir = path.relative(cwd, dir);
```
acburdine: These few lines here will need to be updated to resolve symlinks - for instance the Casper theme is by default a symlink to `./current/content/themes/casper`.

lib/commands/backup.js
```diff
@@ -0,0 +1,200 @@
...
+            title: 'Backing up content folder',
+            task: this.backupContent
+        }, {
+            title: 'Writing backup file',
+            task: this.writeZipFile
+        }],{
+            env: this.system.environment,
+            argv: argv,
+            instance: instance,
+            hook: this.system.hook
```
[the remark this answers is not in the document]
acburdine: Honestly - I'd advise just passing the system object in it's entirety here - that way you don't have to re-bind `this` each time.

vikaspotluri123:
👍 Done @acburdine

lib/commands/backup.js
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    backupExtensions(ctx) {
+        const each = require('lodash/each');
+        const isArray = require('lodash/isArray');
+        const read = fs.readFileSync;
+
+        return ctx.system.hook('backup').then((extensionFiles) => {
+            each(extensionFiles, (filesToAdd) => {
```
acburdine: Just a quick note: you can use `extensionFiles.forEach((filesToAdd) => {})` here, there's not a need for the extra lodash import 😉

test/utils/env.js
```diff
@@ -52,18 +52,18 @@ module.exports = function setupEnv(typeOrDefinition, dir) {
         });
     }
 
-    if (setup.links) {
```
vikaspotluri123: This shouldn't happened. I'll push fixes up tonight or more likely tomorrow.

lib/commands/backup.js
```diff
@@ -0,0 +1,230 @@
...
+
+        // The zip library used (adm-zip) has a known issue where the zip is structurally sound,
+        // but Windows has issues reading / unzipping it. To counteract this, all files are read
+        // and added via the `addFile` method, where the file permissions can be set, seemingly
+        // resolving the issue. See https://github.com/cthackers/adm-zip/issues/171
+        ctx.zipFile = new Zip();
+
+        if (ctx.instance.running()) {
+            ctx.instance.ui.log('Ghost is currently running. Backing up might take longer and slow down your site', 'yellow');
+            ctx.instance.loadRunningEnvironment();
```
acburdine: need to pass `true` to this function - otherwise if it's running in development the NODE_ENV variable won't get set, and the database backup will try to load the wrong config.

lib/commands/backup.js
```diff
@@ -0,0 +1,230 @@
...
+        debug(`Writing zipFile to ${ctx.saveLocation}`);
+        ctx.zipFile.writeZip(ctx.saveLocation);
+        ctx.instance.ui.log(`Wrote backup to ${ctx.saveLocation}`);
+        // clear up memory
+        ctx.zipFile = null;
+    }
+
+    backupDatabase(ctx) {
+        debug('Creating database backup');
+        let exporter;
```
acburdine: ok, so I think I found a way to work around the fact that the database connection isn't terminated by the exporter.

```js
let exporter, db;
try {
  // This is based on cwd, not cli-install location
  db = require(absolutePath('current/core/server/data/db/'));
  exporter = require(absolutePath('current/core/server/data/export/'));
} catch (error) {
  // code omitted for brevity
}
return exporter.doExport().then((database) => {
  // code omitted
  return Promise.fromCallback((cb) => db.knex.destroy(cb));
}); // rest of code omitted for brevity
```

Did a quick test of the above code locally and it worked :)

lib/commands/backup.js
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        return Promise.resolve();
+    }
+
+    backupContent(ctx, dontcare, folder) {
+        folder = folder || './content';
+
+        // path MUST be relative
+        if (folder.indexOf('./') !== 0) {
```
kirrg001: The content folder path in the target config file can be absolute or relative.
If you add `./` to an absolute path, it won't work?

acburdine: @kirrg001 it won't work I think because if the path isn't relative then it will get saved into the zipfile as an absolute path, which can cause issues iirc when unzipping the file.

vikaspotluri123: I completely forgot that the contentPath is configurable 😨 I fixed the issue, but fixing the issue screwed up the zip file. Since everyone seems to be in favor of using `tar.gz` over `zip`, I'm not going to try and fix the new issue.

That being said, I think the content, no matter where it is configured to be for the instance, should be backed up to `{archiveRoot}/content`. When speccing the restore command, we need to be mindful of this, specifically how to react if the content-path (or 1 level up) doesn't exist.

xmrcheck:
Hello, can we know why this feature is on hodl?
Thanks.

acburdine:
@xmrcheck apologies for the delay in getting this feature out. Given that this command is non-trivial, we will need to draft a formal spec for the command first before getting this merged in. Will try to get that done sometime soon.

PR #477, opened 2017-09-14
various dependency-related things

acburdine:


acburdine:
note: couldn't update find-plugins because it (I think accidentally) broke node 4 compatibility - PR to fix it is here: davewasmer/find-plugins#7

acburdine:
well ok then - apparently switching to nyc (and istanbul 1.0) improved coverage by 5% 😂

PR #486, opened 2017-10-07
Strips undesirable characters out of DB name

kieranajp:
If installing Ghost from a folder with a period in its name, for example:

```bash
$ cd example.com
$ ghost install -V
```

The default database name used will contain a period, for example `ghost_production_example.com`.

However, Ghost-CLI will then fail when it attempts to grant permissions to its new user.

This PR strips undesirable characters out of the DB name.

### DIFF
--- a/lib/commands/config/index.js
+++ b/lib/commands/config/index.js
@@ -117,11 +117,13 @@ class ConfigCommand extends Command {
             }
 
             if (!argv.dbname) {
+                const sanitizedDirName = path.basename(process.cwd()).replace(/[^a-zA-Z0-9_]+/g, '_');
                 prompts.push({
                     type: 'input',
                     name: 'dbname',
                     message: 'Enter your Ghost database name:',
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
Looks like you need to update [the test](https://github.com/TryGhost/Ghost-CLI/blob/master/extensions/mysql/test/extension-spec.js#L263) to have quotes

kieranajp:
Yup, no sweat

acburdine:
@kieranajp So I'm okay with this change - but my thought is we should make the default db name one that's easily usable in MySQL - if we let non-standard characters in the default then any manual DB queries also have to be escaped.

kieranajp:
@acburdine Sure, I thought someone might say that.

BTW do I need to bump the version in package.json manually or is that a CI job?

acburdine:
@kieranajp when you say “bump the version” what exactly do you mean?

If you’re taking about bumping it to release a new Ghost-CLI version, no - I’ll take care of making a new release after this is merged.

kieranajp:
Thank _you_!

PR #488, opened 2017-10-08
Node 8

acburdine:
This is not making node 8 officially supported by the CLI yet - the official support still follows https://docs.ghost.org/v1/docs/supported-node-versions

This PR exists to find any places that will be pain points in the future as we seek to enable node 8 support at the end of the month.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - "4"
   - "6"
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
Is this taken care of by #510?

acburdine:
Yup! meant to close this.

Closing in favor of #510

PR #498, opened 2017-10-11
Remove redundant linebreak filter

vikaspotluri123:
The `trim` method removes linebreaks by default so `split('\n')` would do nothing


kevinansfield:
`trim()` only strips line breaks from beginning and end of a string, `split('\n')` will split on any line breaks left inside the string - are you sure it's not necessary?

vikaspotluri123:
Yep, you're right, it is necessary :no_mouth:

vikaspotluri123:
I wasn't thinking straight 😔 the trim method removes _whitespace_ so the _filter_ call is unnecessary - anything that would be falsey from the original string would be removed by trim.

Issue #503, opened 2017-10-18
SPIKE: Investigate knex-migrator improvements


acburdine:
### Summary

One of the most frequent problems that I've seen in both Slack and in the various issues created here is with database migrations (surprise surprise), especially where sqlite is concerned, since it requires native bindings that are specific to the individual version of node.

The most often problem that happens is a user installs ghost-cli, then installs a new version of node via nvm or something similar, and then tries to install Ghost. Database migration fails because knex-migrator's sqlite dependency installed in ghost-cli is for an older version of node, causing database migrations to fail.

There's a couple of things that I think can be done to improve this, which need to be investigated

- Run knex-migrator programmatically rather than as a sub-command

This was actually the case a while back, until we added in the `ghost` system user which necessitated a switch to running it via a subprocess. If we can get it running programmatically, and then run migrations via a subcommand _only_ when we need to run it as the `ghost` system user, this will likely improve usage/error messages/etc.

- Use the knex-migrator version that's installed with Ghost itself and drop the required dep in ghost-cli

This should (hopefully) see significant improvements in the database migration flow. By moving the requirement into Ghost (where knex-migrator is already installed) we accomplish several things:
  - no native dependencies installed with Ghost-CLI
  - removing the "node version is the same that installed ghost-cli" check
  - if any important knex-migrator fixes are shipped with Ghost, Ghost-CLI doesn't need to release a new version in order to get those same fixes

However, this change will need to be verified to ensure that requiring the version installed with Ghost will actually work 😅

acburdine:
Closed by #505

PR #505, opened 2017-10-21
feat(db): remove knex-migrator dep, use one installed with Ghost

acburdine:
refs #503
- use knex-migrator version installed with Ghost itself
- remove global knex-migrator dependency
- remove "same node version" check

### DIFF
--- a/lib/commands/doctor/checks/install.js
+++ b/lib/commands/doctor/checks/install.js
@@ -37,16 +37,6 @@ const tasks = {
         });
     },
     nodeVersion: function nodeVersion(ctx) {
-        const globalBin = execa.shellSync('npm bin -g', {preferLocal: false}).stdout;
-
-        if (process.argv[1] !== path.join(__dirname, '../../../../bin/ghost') && !process.argv[1].startsWith(globalBin)) {
-            return Promise.reject(new errors.SystemError(
-                `The version of Ghost-CLI you are running was not installed with this version of Node.${eol}` +
-                `This means there are likely two versions of Node running on your system, please ensure${eol}` +
-                'that you are only running one global version of Node before continuing.'
-            ));
-        }
-
         if (process.env.GHOST_NODE_VERSION_CHECK !== 'false' &&
             !semver.satisfies(process.versions.node, cliPackage.engines.node)) {
             return Promise.reject(new errors.SystemError(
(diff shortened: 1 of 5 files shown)


kirrg001:
That would be a great improvement 👍 

As soon as we merge the v3 of knex-migrator into Ghost (see https://github.com/TryGhost/Ghost/pull/9199), i am concerned that people will run into problems because of different versions.

Is this update something we could release asap? Do you need any help?

kirrg001:
Is there a concept of forcing people to update to the latest CLI version?

acburdine:
@kirrg001 To the first question: yes - I can get this released today or tomorrow. The second question: no, there isn't a way currently to force people to upgrade to the latest Ghost-CLI version. However, I have an idea as to how to make that work, will create a PR for it asap.

lib/commands/doctor/checks/install.js
```diff
@@ -37,16 +37,6 @@ const tasks = {
         });
     },
     nodeVersion: function nodeVersion(ctx) {
-        const globalBin = execa.shellSync('npm bin -g', {preferLocal: false}).stdout;
```
kirrg001: Can you explain why this removal is part of this commit? 🤔

acburdine: The reason why we needed that check in the first place was because sqlite is node version specific.

Because we're using the knex-migrator installed with Ghost itself (and hence the sqlite version installed there), this check isn't necessary anymore.

lib/tasks/migrate.js
```diff
@@ -24,12 +24,12 @@ module.exports = function runMigrations(context) {
...
     // we should run sudo, otherwise run normally
     if (shouldUseGhostUser(contentDir)) {
-        const knexMigratorPath = path.resolve(__dirname, '../../node_modules/.bin/knex-migrator-migrate');
+        const knexMigratorPath = path.resolve(context.instance.dir, 'current/node_modules/.bin/knex-migrator-migrate');
         knexMigratorPromise = context.ui.sudo(`${knexMigratorPath} ${args.join(' ')}`, {sudoArgs: '-E -u ghost'});
     } else {
         knexMigratorPromise = execa('knex-migrator-migrate', args, {
             preferLocal: true,
-            localDir: __dirname
+            localDir: path.join(context.instance.dir, 'current')
```
kirrg001: Short understanding question. `execa` extends the PATH and so it's able to execute the local installation inside `current/node_modules`, correct?

lib/commands/doctor/checks/install.js
kirrg001: Oh i see, thanks 👍

lib/tasks/migrate.js
rv-aburdine: yup

PR #507, opened 2017-10-24
acme.sh usage fixes

acburdine:
this _should_ in theory fix our acme.sh issues, although it needs testing on ubuntu to confirm.

Will test tomorrow to make sure it's working correctly, then merge + release :)

### DIFF
--- a/extensions/nginx/index.js
+++ b/extensions/nginx/index.js
@@ -238,7 +238,7 @@ class NginxExtension extends cli.Extension {
     }
 
     restartNginx() {
-        return this.ui.sudo('service nginx restart')
+        return this.ui.sudo('nginx -s reload')
             .catch((error) => Promise.reject(new cli.errors.ProcessError(error)));
     }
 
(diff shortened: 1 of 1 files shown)


extensions/nginx/index.js
```diff
@@ -131,28 +131,48 @@ class NginxExtension extends cli.Extension {
...
+                    // It's not an error we expect might happen, throw a ProcessError instead.
+                    return Promise.reject(new cli.errors.ProcessError(error));
                 });
             }
         }, {
             title: 'Generating Encryption Key (may take a few minutes)',
             skip: (ctx) => ctx.dnsfail || fs.existsSync(dhparamFile),
             task: () => {
-                return this.ui.sudo(`openssl dhparam -out ${dhparamFile} 2048 > /dev/null`)
+                return this.ui.sudo(`openssl dhparam -out ${dhparamFile} 2048`)
```
ErisDS: This change looks like it doesn't belong here?

acburdine: yeah - i'll fix it in a separate PR. That redirect actually does nothing at the moment 😁

PR #510, opened 2017-10-26
Updated package.json to add support for Node v8.8+

tuesdaywednesday:
To comply with the following statement found in [supported-node-versions](https://docs.ghost.org/docs/supported-node-versions).

`Ghost 1.0.0 currently supports Node versions 8.8+, 6.9+ and 4.5+ only.`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - "4"
   - "6"
+  - "8"
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


acburdine:
Hey @tuesdaywednesday, thanks for the change!

Would you mind also adding node 8 to the .travis.yml - so we can make sure tests pass on node 8?

Thanks!

tuesdaywednesday:
@acburdine Sure! Done just that!

dennisroche:
Any chance we can get a release published with this fix?

acburdine:
@dennisroche yup! didn't release last night because it was super late and had to make a few other changes. Will get that release out today 😄

Issue #511, opened 2017-10-26
Creating new mysql user errored with message: ER_NOT_VALID_PASSWORD: Your password does not satisfy the current policy requirements


imzhengfei:
Welcome to Ghost-CLI's GitHub repo! 👋🎉

Do you need help or have a question? Please come chat in Slack: https://slack.ghost.org/ 👫.

### This issue is a

* [x] Bug Report
* [ ] Feature Request

### Summary
It seems the random password for MySQL user is too simple.

> [root@web blog]# ghost setup mysql
> ✖ Setting up "ghost" mysql user
> A SystemError occurred.
> 
> Message: Creating new mysql user errored with message: ER_NOT_VALID_PASSWORD: Your > password does not satisfy the current policy requirements
> 
> Debug Information:
>     Node Version: v6.11.5
>     Ghost-CLI Version: 1.1.3
>     Environment: production
>     Command: 'ghost setup mysql'
> 
> Please refer to https://docs.ghost.org/v1/docs/troubleshooting#section-cli-errors for troubleshooting.
> [root@web blog]# mysqld -V
> mysqld  Ver 5.7.20 for Linux on x86_64 (MySQL Community Server (GPL))

allthedrones:
I also encountered this issue, with the MySQL Password Validation Plugin (which MySQL now prompts for when running the included hardening routine `mysql_secure_installation`), configured to the MEDIUM or STRONG settings. 

Most likely RC is password shorter than 8 chars or not meeting complexity requirements (mixed case, 1+ numerals, 1+ special char).

Failing password config on MySQL:

```
mysql> SHOW VARIABLES LIKE 'validate_password%';
+--------------------------------------+--------+
| Variable_name                        | Value  |
+--------------------------------------+--------+
| validate_password_check_user_name    | OFF    |
| validate_password_dictionary_file    |        |
| validate_password_length             | 8      |
| validate_password_mixed_case_count   | 1      |
| validate_password_number_count       | 1      |
| validate_password_policy             | MEDIUM |
| validate_password_special_char_count | 1      |
+--------------------------------------+--------+
```

acburdine:
@imzhengfei this should I think be a fairly small/easy fix - so if you want to create a PR for it feel free :) Otherwise I'll get to this when I have some time.

imzhengfei:
@acburdine I know little about node, I will try. Hope I can do it. :)

gompa:
[...]

PR #517, opened 2017-11-02
fix(mysql): improve random(with LCG) password for mysql user

imzhengfei:
fix: #511

### DIFF
--- a/extensions/mysql/index.js
+++ b/extensions/mysql/index.js
@@ -2,7 +2,6 @@
 
 const Promise = require('bluebird');
 const mysql = require('mysql');
-const crypto = require('crypto');
 const omit = require('lodash/omit');
 const cli = require('../../lib');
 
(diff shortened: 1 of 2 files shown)


imzhengfei:
Use Linear Congruential Generator to generate a random number between 0 and 1, in stead of Math.random().

extensions/mysql/index.js
```diff
@@ -77,7 +76,74 @@ class MySQLExtension extends cli.Extension {
     }
 
     createUser(ctx, dbconfig) {
-        const randomPassword = crypto.randomBytes(10).toString('hex');
+        // Generate random MySQL password
+        const generateRandomPassword = () => {
+            /**
+             * LCG Random
+             * @example cosnt random = (lcgRandom())
```
vikaspotluri123: Was this supposed to say `const` instead of `cosnt`?

imzhengfei: Oh! Sorry! Just a spelling error.

imzhengfei:
fixed a spelling error

imzhengfei:
@acburdine I am not sure, do you mean MySQLExtension class ?

acburdine:
@imzhengfei my apologies, yes that's what I meant.

However, as I commented [here](https://github.com/TryGhost/Ghost-CLI/pull/534#issuecomment-344140943), there's another issue in that the initial `create user` command is what actually is triggering the insecure password error - so we need to find a way around that as well.

acburdine:
Closing in favor of #546 -> ultimately it required other changes than this. @gompa's approach in #534 was a bit better than this I think too - it was less code to maintain + the library used in that PR uses crypto.randomBytes rather than something based off of a technically predictable seed.

Issue #496, opened 2017-10-10
Can't install multiples instances "Port '2368' is in use"


treyssatvincent:
### This issue is a

- [x] Bug Report
* [ ] Feature Request

### Summary

If this is a bug, double-check and make sure there's not already an existing issue for it, then provide a quick summary of what error is occurring.

### Steps to Reproduce (for a bug report)

1. run ``ghost install`` on /var/www/exampleone
2. run ``ghost install` on /var/www/exampletwo

> ? Enter your Ghost database name: ghost_production_fixmystreet
> ✖ Configuring Ghost
> A ConfigError occurred.
> 
> Error detected in the undefined configuration.
> 
> Message: Port '2368' is in use.
> Configuration Key(s): server.port
> Current Value(s): 2368
> 
> Help: Run `ghost config <key> <new value>` for each key to fix the issue.

Debug Information:
    Node Version: 
    Ghost-CLI Version: 1.1.3
    Environment: production
    Command: 'ghost install'

### Technical details (will be automatically output by Ghost-CLI if an error occurs):

* OS: Ubuntu 16.04
* Node Version: v6.11.4
* Ghost-CLI Version: 1.1.3
* Environment: production
* Command: ghost install

ErisDS:
I can confirm this bug, I also ran into it today.

TomChantler:
This is odd as we had an issue a while back with port numbers limiting multiple installations and which was described in #451 and fixed in #452. I wonder why it's broken again? I'll try to take a look.

lmuzquiz:
I also ran into this issue today. 

This is what i did: 

mkdir /var/www/lmuzquiz/blog
cd /var/www/lmuzquiz/blog
ghost install

Then after installing ghost successfully, i  intentionally removed the whole blog folder by running: 

rm -rf /var/www/lmuzquiz/blog

Created the blog folder again, ran ghost install, and the installation process begins smoothly but after i answer this question: 

Enter your Ghost database name: my_ghost_db

That is when i get the error: 

`A ConfigError occurred.

Error detected in the undefined configuration.

Message: Port '2368' is in use.
Configuration Key(s): server.port
Current Value(s): 2368

Help: Run `ghost config <key> <new value>` for each key to fix the issue.


Debug Information:
    Node Version: v6.11.4
    Ghost-CLI Version: 1.1.3
    Environment: production
    Command: 'ghost install'
`

TomChantler:
[...]

PR #526, opened 2017-11-10
fix(config): ports work again

### LINKED ISSUE #519: Install using http://server.com:port overwrites port# with 2368
When installing and using a URL with a port in it, the port is seen but the value is replaced with 2368 in both the URL and the `port` setting in the config.  

Steps performed:
----
* Create directory `/opt/ghost-v1/`, and `chown ghost:ghost` that thing
* Run ghost install in the directory as root
* Enter "https://somesite.co:4000" as the url

Results:
----
* Config contains `url = https://somesite.co:4000`
* Config contains `port = 2368`

Desired Results:
----
Not really sure what is the right/easy/common way to go.  It would be nice to have 2 settings, one for port and one for url, though I could just edit the config file after and remove the `:4000` from the blog url, since its fronted by nginx.  But if its there they should at least be the same.

Install Output:
----
```bash
root@linode:/opt/ghost-v1# ghost install --db=sqlite3
✔ Checking system Node.js version
...
✔ Downloading and installing Ghost v1.16.2
✔ Finishing install process

? Enter your blog URL: http://somesite.co:4000
✔ Configuring Ghost
✔ Setting up instance
...
? Do you wish to set up Systemd? No
ℹ Setting up Systemd [skipped]
Running sudo command: /usr/lib/node_modules/ghost-cli/node_modules/.bin/knex-migrator-migrate --init --mgpath /opt/ghost-v1/current
...
```

Resulted in this config (unrelated sections omitted)
```
root@linode:/opt/ghost-v1# cat config.production.json
{
  "url": "http://somesite.co:2368/",
  "server": {
    "port": 2368,
    "host": "127.0.0.1"
  },
 ....
```

kirrg001:
closes #519

- `ghost install` -> uses the next available port (starts with 2368)
- `ghost install --url http://localhost:4000` (uses 4000)
- `ghost install --url http://localhost:4000 --port 2390` (uses 2390 and saves localhost:2390)

### DIFF
--- a/lib/commands/config/advanced.js
+++ b/lib/commands/config/advanced.js
@@ -55,7 +55,6 @@ module.exports = {
             const port = parseInt(url.parse(config.get('url')).port || BASE_PORT);
             return portfinder.getPortPromise({port: port});
         },
-        default: 2368,
         type: 'number',
         group: 'Ghost Options:'
     },


kirrg001:
refs #496

laurosilvaio:
@kirrg001 Can you be more specific on how you solved the problem. Are we suppose to run `ghost install --url http://localhost:4000 --port 2390`?

kirrg001:
@laurosilvaio Depends on the use case, but all the cases are described in the PR description.

> ghost install --url http://localhost:4000 --port 2390 (uses 2390 and saves localhost:2390)

PR #530, opened 2017-11-11
fix(extension): ensure require(ghost-cli) returns the root instance

acburdine:
Not merging this yet - just had an idea on how to fix the issue and wanted to test it out.


acburdine:
Tested this on ubuntu - it works :)

Issue #396, opened 2017-07-28
mysql user setup sometimes has issues with creating passwords


acburdine:
As reported by @Penderworth over in #378:

<details>
  <summary>Error Log</summary>

```
Debug Information:
    Node Version: v6.5.0
    Ghost-CLI Version: 1.0.0
    Environment: production
    Command: 'ghost install'
An error occurred.
Message: 'Command failed: /bin/sh -c sudo -E -u ghost /usr/local/lib/node_modules/ghost-cli/node_modules/.bin/knex-migrator-migrate --init --mgpath /var/www/ghost/current
[2017-07-27 10:08:56] ERROR

NAME: InternalServerError
CODE: ER_NOT_SUPPORTED_AUTH_MODE
MESSAGE: The server has encountered an error.

level:normal

InternalServerError: The server has encountered an error.
    at new KnexMigrateError (/usr/local/lib/node_modules/ghost-cli/node_modules/knex-migrator/lib/errors.js:7:26)
    at /usr/local/lib/node_modules/ghost-cli/node_modules/knex-migrator/lib/database.js:58:19
    at tryCatcher (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/promise.js:689:18)
    at Async._drainQueue (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/async.js:133:16)
    at Async._drainQueues (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.Async.drainQueues (/usr/local/lib/node_modules/ghost-cli/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:574:20)
    at tryOnImmediate (timers.js:554:5)
    at processImmediate [as _immediateCallback] (timers.js:533:5)

[...]

PR #534, opened 2017-11-14
fix(mysql): improve password compatibility mysql user

gompa:
improve password compatibility with mysql password requirements
by adding the generate-password node module

### DIFF
--- a/extensions/mysql/index.js
+++ b/extensions/mysql/index.js
@@ -2,9 +2,9 @@
 
 const Promise = require('bluebird');
 const mysql = require('mysql');
-const crypto = require('crypto');
 const omit = require('lodash/omit');
 const cli = require('../../lib');
+const generator = require('generate-password');
 
 class MySQLExtension extends cli.Extension {
     setup(cmd, argv) {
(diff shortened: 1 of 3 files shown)


acburdine:
hmm...now that I'm looking at this regex it might actually have issues 😕 I'll test it out though and see if it works/doesn't.

vikaspotluri123:
@acburdine imho the regex can be simplified to something like `^SET PASSWORD FOR 'ghost-[0-9]{1,4}'@'localhost' = PASSWORD\('.{10,}'\);$` (or can be condensed even more) - the tests are verifying queries and retries rather than a strong password is generated.

vikaspotluri123:
Also sidenote: I think this is similar to (if not the same idea as) #517

acburdine:
yeah it's a similar idea - we had a convo on Slack about this approach as well.

However, I figured out that the underlying issue wasn't actually the password itself (although that is a factor). The first issue we have to solve is that the `CREATE USER 'ghost-xx'@'localhost' IDENTIFIED WITH mysql_native_password` _also_ throws an error, before we even get to the secure password part.

That mysql_native_password thing was added to fix #396 - we need to ensure that whatever solution doesn't make that issue happen again.

acburdine:
@gompa thanks for the fix! I went ahead and cherry-picked this commit and applied it to #546 -> I had to make some other changes alongside this one.

PR #535, opened 2017-11-14
feat(ui): improve update notifications

### LINKED ISSUE #525: Update notifications for Ghost-CLI
Related to #505, TryGhost/Ghost#9199, #495 

Currently we have an issue in the CLI in that there is no serious prompting to upgrade the CLI itself. There is currently an upgrade notification, but that happens _after_ a ghost command is run, and doesn't specify that it is a _required_ upgrade, just that there is a new version available.

We need to implement a way to either require or strongly suggest that a user update the version of Ghost-CLI. There are two scenarios in which I think this is required:

1. A critical bugfix in the CLI itself, unrelated to changes in Ghost. (example: #495)

In this situation, an issue was discovered with the nginx setup, and a fix was released in 1.2.0. However, people would not get that fix unless they updated to the latest version of the CLI. We need to be able to tell these users that they need to upgrade in order to get these fixes.

2. A breaking change in Ghost itself that requires a CLI update. (example, TryGhost/Ghost#9199)

In this scenario, Ghost is changing the migration files themselves due to breaking changes in the knex-migrator dependency. Because the current version of the CLI depends on an old version of knex-migrator, we would need to prompt the user to update Ghost-CLI prior to installing the new version of Ghost.

There are two checks that we can add to fix these issues:

1. Check for a new Ghost-CLI version when `any` ghost command is run, and require the user to either update first, or continue without updating. Something like this:

```sh
$ ghost update # or ghost ls, ghost install, etc
Warning: a new version of Ghost-CLI is available that contains some important fixes.
You should update your CLI version before running this command.
(Run `npm install -g ghost-cli@latest` to update)
Continue anyways? [y/N]:
```

2. Add a semver range in the package.json of Ghost itself that specifies what versions of the CLI it is compatible with.

Ghost-CLI would check this prior to actually downloading & updating to the latest version of Ghost. If the current CLI version is older than the version that Ghost's package.json says its compatible with, it will error out with a message saying ghost-cli needs to be updated.

acburdine:
closes #525
- make update check more prominent

TODO: add tests

### DIFF
--- a/bin/ghost
+++ b/bin/ghost
@@ -4,11 +4,22 @@
 // provide a title to the process
 process.title = 'ghost';
 
-// Check for CLI updates if not running `ghost run`
-if (process.argv.length > 2 && process.argv[2] !== 'run') {
-    require('../lib/utils/update-notifier');
+function startup() {
+    const yargs = require('yargs');
+    const bootstrap = require('../lib/bootstrap');
+    bootstrap.run(process.argv.slice(2), yargs);
 }
 
-const yargs = require('yargs');
-const bootstrap = require('../lib/bootstrap');
-bootstrap.run(process.argv.slice(2), yargs);
+// Check for CLI updates if not running `ghost run`
+if (process.argv.length > 2 && process.argv[2] !== 'run') {
+    const updateCheck = require('../lib/utils/update-check');
+    updateCheck().then(startup).catch((error) => {
+        if (error) {
+            console.error(error);
+        }
+
+        process.exit(1);
+    });
+} else {
+    startup();
+}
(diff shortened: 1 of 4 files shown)


bin/ghost
```diff
@@ -4,11 +4,22 @@
...
+    const bootstrap = require('../lib/bootstrap');
+    bootstrap.run(process.argv.slice(2), yargs);
+}
+
 // Check for CLI updates if not running `ghost run`
 if (process.argv.length > 2 && process.argv[2] !== 'run') {
-    require('../lib/utils/update-notifier');
-}
+    const updateCheck = require('../lib/utils/update-check');
+    updateCheck().then(() => startup()).catch((error) => {
```
vikaspotluri123: can this not be `updateCheck().then(startup)...`?

acburdine: yup - it definitely can be :)

Issue #495, opened 2017-10-10
Acme.sh certificate renewal is not complete


ErisDS:
I'm having problems with sites going offline with certificate errors. Acme.sh is correctly renewing the certificates, but nginx must be restarted afterwards, and I don't think that has been configured.

I'm not sure why we haven't had a tonne more reports of this!

Refs: 

- https://github.com/Neilpang/acme.sh#3-install-the-issued-cert-to-apachenginx-etc
- https://www.rmedgar.com/blog/using-acme-sh-with-nginx

**Update from @ErisDS 23rd Oct:**

We're starting to see more people affected by this. To aid search-ability this manifests itself as `NET::ERR_CERT_DATE_INVALID` errors on your site.

The certificate _has_ renewed, but nginx needs to be restarted to pick up the new certs:

The solution is to run `sudo systemctl restart nginx` on your box.

--- 

## Task list 

(Taken from my comment below)

- [x] 1. Build out the new version
- [ ] 2. Write the documentation for how to manually migrate your site to the new version
     - My expectation is the simple version would be "run `ghost setup:ssl`, then delete X files" or similar, and then there'd be an extended version explaining the exact steps for anyone with a customised system.
- [x] 3. Release the new version as a minor?

At this point, all new sites are safe, all old ones _can_ migrate.

- [ ] 4. Write a check that can determine if a blog is using the old ssl setup, make it as robust and simple as possible. E.g. use `fs.exists`.
- [ ] 5. Add a message to as many places as possible shouting about the need to update, linking to the docs.
- [ ] 6. Release as a patch update

At this point, all old blogs will learn about the need to migrate. They will be able but it's not easy.

- [ ] 7. Add a new command: `ghost migrate:ssl` or something equally relatively sensible and future-proof.
- [ ] 8. Make that command do the necessary migration steps 
      - e.g. it runs `ghost setup:ssl` and then deletes the old files, or maybe does a more advanced set of steps to be more compatible?
- [ ] 9. Update the docs with the shiny new command
- [ ] 10. _Maybe_ update the messages every where with the new command?

At this point, all old blogs have an easy way to migrate.

- [ ] 11. Deal with any fall out, e.g. bugs in the check, or improvements to the migrate command to ensure these are robust.
- [ ] 12. Implement a system to auto-migrate

benaubin:
[...]

PR #536, opened 2017-11-14
CLI Migrations

acburdine:
refs #495 

TODOS:
- [x] finish ssl migration stuff
- [x] add migrate command to run migrations explicitly
- [x] add tests
- [x] double check manual test & re-order commits

### DIFF
--- a/extensions/nginx/acme.js
+++ b/extensions/nginx/acme.js
@@ -1,0 +1,92 @@
+'use strict';
+const fs = require('fs-extra');
+const os = require('os');
+const got = require('got');
+const path = require('path');
+const download = require('download');
+
+const cli = require('../../lib');
+
+function isInstalled() {
+    return fs.existsSync('/etc/letsencrypt/acme.sh');
+}
+
+function install(ui, task) {
+    if (isInstalled()) {
+        return task.skip('acme.sh is already installed');
+    }
+
+    const acmeTmpDir = path.join(os.tmpdir(), 'acme.sh');
+    const acmeApiUrl = 'https://api.github.com/repos/Neilpang/acme.sh/releases/latest';
+
+    ui.logVerbose('ssl: creating /etc/letsencrypt directory', 'green');
+
+    // acme.sh creates the directory without global read permissions, so we need to make
+    // sure it has global read permissions first
+    return ui.sudo('mkdir -p /etc/letsencrypt').then(() => {
+        ui.logVerbose('ssl: downloading acme.sh to temporary directory', 'green');
+        return fs.emptyDir(acmeTmpDir)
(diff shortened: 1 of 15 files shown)


acburdine:
code works, just need to add tests now

extensions/nginx/acme.js
```diff
@@ -0,0 +1,91 @@
...
+            ));
+        }
+
+        // It's not an error we expect might happen, throw a ProcessError instead.
+        return Promise.reject(new cli.errors.ProcessError(error));
+    });
+}
+
+function remove(domain, ui) {
+    const acmeHome = path.join(os.homedir(), '.acme.sh');
```
vikaspotluri123: Isn't the ACME install location `/etc/letsencrypt`?

lib/commands/update.js
```diff
@@ -75,7 +76,27 @@ class UpdateCommand extends Command {
             task: this.removeOldVersions
         }];
 
-        return this.ui.run(() => this.version(context), 'Checking for latest Ghost version').then((result) => {
+        return this.ui.run(
```
vikaspotluri123: This looks really similar to the migrations task - is it possible to run the migration command, similar to the way [install calls setup](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/install.js#L85)?

extensions/nginx/acme.js
acburdine: not in this case, no. We're migrating from the original acme install location to the new location, which means we want to use the old location here.

I can update the variable name to be more clear though

lib/commands/update.js
acburdine: good idea!

extensions/nginx/acme.js
vikaspotluri123: Got it. Should the nginx uninstall hook be calling something similar to this to delete the certificate from the system? This might be a separate issue 🤔

acburdine: hmm yeah that's not a bad idea

acburdine:
tests added 🎉 will double check this again in the morning before merging :)

acburdine:
@kirrg001 if you want to run any tests yourself, there are three different scenarios I want to test manually:

- [ ] setting up ssl with ghost < 1.2.0, then running ghost migrate with this PR
- [ ] setting up ssl with ghost < 1.2.0, then following [these instructions](https://docs.ghost.org/v1/docs/upgrade#section-upgrading-from-1-1-3-to-1-2-0) on manually fixing ssl, then running ghost migrate with this pr
- [ ] setting up ssl with this PR from scratch (to ensure we didn't break anything 😁)

kirrg001:
@acburdine will do 👍

Issue #541, opened 2017-11-16
Road to 💯 - Tests


vikaspotluri123:
Here's an abstract list of things the cli does, and how well they're tested (via unit tests)

### Extensions: 
- [x] MySQL
- [x] Nginx (#523, #542)
- [x] Systemd (#544)

### Commands: 
- [x] buster
- [x] install
- [x] log (#545)
- [x] ls
- [x] migrate
- [x] restart
- [x] run
- [x] setup (#559)
- [x] start (#560)
- [x] stop (#582)
- [x] uninstall (#585)
- [x] update 
- [x] version

### Tasks:
- [x] ensure-structure
- [x] linux
- [x] migrate
- [ ] yarn-install (#586)

### Utils:
- [x] config
- [x] find-extensions
- [x] needed-migrations
- [x] resolve-version
- [x] update-check
- [x] use-ghost-user
- [x] version-from-zip
- [ ] yarn (#586)

### Misc:
- [x] Remove spec files from coverage reports (#543)
- [x] bootstrap

AileenCGN:
When I run the tests locally, there's always one test that fails, but it's fine in Travis:

 ```
 1) Unit: Extensions > Nginx
       migrations hook
         [before 1.2.0]
           Doesn't skip if acme.sh exists:

      AssertionError: expected true to be false
      + expected - actual

      -true
      +false

      at Context.<anonymous> (extensions/nginx/test/extension-spec.js:73:51)
```

vikaspotluri123:
@AileenCGN I _think_ it's because we haven't stubbed `os`; I'm thinking `os.platform() !== linux` evaluates to `true` locally for you.

I haven't hit this because I test with [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux)  which is technically a linux platform

PR #543, opened 2017-11-17
Don't check coverage on test files

vikaspotluri123:
Currently, test files (`*-spec.js`) are tested for coverage in nyc. This isn't necessarily something that should be done, since the goal of tests is to execute core files, and certain areas _should not_ be run.

Ref: #541 (kind of)

Full documentation available in [nyc readme](https://github.com/istanbuljs/nyc#excluding-files)

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "coverage": "nyc report --reporter=text-lcov | coveralls",
     "lint": "eslint bin lib test extensions",
     "release": "standard-version",
-    "test:unit": "nyc --reporter=html --reporter=text mocha --recursive test/unit extensions/**/test",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -34,6 +34,12 @@
     "test:all": "yarn run test:unit && yarn run test:acceptance",
     "test": "yarn run lint && yarn run test:all"
   },
+  "nyc": {
+    "exclude": [
+      "**/*-spec.js",
+      "test"
```
vikaspotluri123: I'm not sure why I needed to add this, but when I was running unit tests locally, the test/utils folder decided to pop up

PR #544, opened 2017-11-17
chore(tests): add systemd process manager tests

vikaspotluri123:
I feel like there's some repetition, so if you have any suggestions, feel free to comment!

ref: #541


acburdine:
Nah, I don't think it's too bad. We _could_ abstract those tests out, but then we'd have to split them out in the future if/when we ever change the behavior of one function.

PR #547, opened 2017-11-18
fix(run): make `ghost run` responsible for outputting running env

### LINKED ISSUE #463: Make `ghost run` be responsible for outputting running environment
There have been several issues raised recently (#461, #449, #431, #430) where the CLI loses track of running blog state when using systemd. The root cause of this is related to how the CLI keeps track of instance running environment.

Currently, `ghost start` starts Ghost using the configured process manager for the particular environment, and after it has been started sticks a value into the `.ghost-cli` file at the root of the install that notes what environment Ghost is currently running in. This value is used by the CLI as the source of truth about whether or not a blog is running. If the value does not exist in the .ghost-cli file the blog is assumed to not be running; if it exists, the configured process manager for the running environment is queried to ensure that the blog is actually still running. This approach has issues with systemd in a couple of cases:

- you run `ghost stop`, then reboot the server. Because the systemd service has been configured to start on boot, the ghost instance is started on boot but no value is output to the `.ghost-cli` file because ghost was started automatically, not using `ghost start`
- you use `systemctl start/stop ghost_<servicename>` to start/stop Ghost rather than `ghost start`/`ghost stop`. This causes the same issue of the running environment not being tracked.

To fix this, we need to give responsibility for updating running environment to the `ghost run` command (which both the local process manager _and_ the systemd process manager use under the hood). However, we can't make `ghost run` modify the .ghost-cli file as `ghost start` does right now - because systemd runs `ghost run` as the ghost system user, it won't have permissions to edit that file. Instead, `ghost run` will output a `.environment` file into the `content` directory of the ghost install (which the ghost system user _does_ have write access to) and then the CLI will use that file to determine what environment the CLI is currently running in.

acburdine:
closes #463 [ci skip]
- output a .environment file into the content folder using `ghost run`, is cleaned up when ghost is stopped

todo:
- [ ] fix tests
- [ ] manual test the 💩 out of this

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -1,10 +1,11 @@
 'use strict';
+const fs = require('fs-extra');
+const path = require('path');
 const spawn = require('child_process').spawn;
 const Command = require('../command');
 
 class RunCommand extends Command {
     run() {
-        const path = require('path');
         const shouldUseGhostUser = require('../utils/use-ghost-user');
 
         // If the user is running this command directly, output a little note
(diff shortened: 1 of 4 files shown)


acburdine:
Going with a different approach to solve #463, but will leave the branch around just in case.

PR #556, opened 2017-12-01
fix(ssl-migration): make the checks for migrating ssl more robust

### LINKED ISSUE #552: [Migrations] Failure when a domain wasn't setup via CLI
# Problem

There are 2 instances - one had ssl setup through the CLI pre 1.2.0 and the other had it setup elsewhere. The CLI is upgraded to 1.3.0, and `ghost update` is run for the CLI-setup domain. Everything works (🎉). `ghost update` is run for the no-CLI-setup domain. It fails:

<details><summary>no-CLI-ssl Log</summary><pre>
user@user-ghost:/var/www/other$ sudo ghost upgrade
✔ Checking for available migrations
Running sudo command: /etc/letsencrypt/acme.sh --issue --home /etc/letsencrypt --domain example.com --webroot /var/www/other/system/nginx-root --reloadcmd "nginx -s reload" --accountemail user@example.com
Running sudo command: nginx -s reload
Running sudo command: /home/user/.acme.sh/acme.sh --remove --home /home/user/.acme.sh --domain example.com
✖ Migrating SSL certs
A ProcessError occurred.
Error occurred running command: '/bin/sh -c sudo -S -p '#node-sudo-passwd#'  /home/uesr/.acme.sh/acme.sh --remove --home /home/user/.acme.sh --domain example.com'
Exit code: 1
Debug Information:
    Node Version: v6.12.0
    Ghost-CLI Version: 1.3.0
    Environment: production
    Command: 'ghost upgrade'</pre></details>
<br/>

Note: I later realized the issue occurred because a different user installed the second instance - ssl _was_ setup with the CLI, but the certificate wasn't in the same acme path as the other one. This issue now impacts even less of the user base, but is something that should still be fixed.

The solution @acburdine and I discussed was checking if the domain exists in [skip check](https://github.com/TryGhost/Ghost-CLI/blob/master/extensions/nginx/index.js#L21)  (i.e. `!fs.existsSync(path.join(os.homedir(), '.acme.sh',  domain))` as well as catch the error from acme, skipping if this specific case is the issue. I _think_ the log contains all the info necessary to fix this 😃 

# Temporary Solution

If you're facing this issue and are 100% sure you don't need to migrate, edit the `.ghost-cli` file and update the line that says `"cli-version": "{something before 1.2.0}"` to `"cli-version": "1.3.0"`. Be warned: if you did need to run migrations, and you don't, nginx won't reload the SSL certificate when your certificate renews, meaning it will be sending an expired certificate soon after. Only do this if you are 100% sure you don't need to run migrations. Make sure you try to run migrations as the user that installed the instance.

acburdine:
closes #552
- skip ssl migration in a couple more cases

TODO:
- [x] fix tests
- manual test all the things
  - [x] test 1/3
  - [x] test 2/3
  - [x] test 3/3

### DIFF
--- a/extensions/nginx/migrations.js
+++ b/extensions/nginx/migrations.js
@@ -14,11 +14,26 @@ function migrateSSL(ctx, migrateTask) {
     const confFile = path.join(ctx.instance.dir, 'system', 'files', `${parsedUrl.hostname}-ssl.conf`);
     const rootPath = path.resolve(ctx.instance.dir, 'system', 'nginx-root');
 
+    // Case to skip 1: SSL config for nginx does not exist
     if (!fs.existsSync(confFile)) {
         return migrateTask.skip('SSL config has not been set up for this domain');
     }
 
     const originalAcmePath = path.join(os.homedir(), '.acme.sh');
+    const originalCertFolder = path.join(originalAcmePath, parsedUrl.hostname);
+
+    // Case to skip 2: SSL cert doesn't exist in the original location for this domain
+    if (!fs.existsSync(originalCertFolder)) {
+        return migrateTask.skip('SSL cert does not exist for this domain');
+    }
+
+    const confFileContents = fs.readFileSync(confFile, {encoding: 'utf8'});
+    const certCheck = new RegExp(`ssl_certificate ${originalCertFolder}/fullchain.cer;`)
+
+    // Case to skip 3: SSL conf does not contain a cert config using the old LE cert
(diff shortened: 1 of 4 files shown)


ErisDS:
I've taken a quick read over the code & looks good.

acburdine:
Tested three separate use-cases:

1. no acme cert has been generated for the domain, but an ssl conf exists
2. acme cert exists but is not currently being used
3. normal migration still works

All three were good 👍

PR #558, opened 2017-12-02
Make update-check honor no-prompt flag

### LINKED ISSUE #563: [BUG] updateCheck disallow non-interactive setup
https://github.com/TryGhost/Ghost-CLI/blob/ff5e4ad9c30e75a10134741f96de31642123bce8/bin/ghost#L14

From ghot-cli 1.3.0 on wards, there is no way to proceed with unattended installations (`--no-prompt`) when a newer version of ghos-cli is detected. It could be useful in some circumstances but, under different ones, it should be avoidable.

I.E. when trying to install you need to pass `ghost` a lot of parameters if using in a non-interactive way `setup ........` but it becomes interactive because `inquirer` launch a question after the `updateCheck`. This is not consistent.

Related: https://github.com/TryGhost/Ghost-CLI/issues/525

vikaspotluri123:
closes #563

### DIFF
--- a/lib/utils/update-check.js
+++ b/lib/utils/update-check.js
@@ -28,21 +28,21 @@ module.exports = function updateCheck() {
 
                 // Yargs hasn't run yet
                 if (includes(process.argv, '--no-prompt')) {
-                    reject('Requested no prompting. Exiting...');
-                } else {
-                    inquirer.prompt({
-                        type: 'confirm',
-                        message: 'Continue without upgrading?',
-                        default: false,
-                        name: 'yes'
-                    }).then((answers) => {
-                        if (answers.yes) {
-                            return resolve();
-                        }
-
-                        reject();
-                    }).catch(reject);
+                    return reject('Requested no prompting. Exiting...');
                 }
+
+                inquirer.prompt({
+                    type: 'confirm',
+                    message: 'Continue without upgrading?',
+                    default: false,
+                    name: 'yes'
+                }).then((answers) => {
+                    if (answers.yes) {
+                        return resolve();
+                    }
+
+                    reject();
+                }).catch(reject);
             }
         });
     });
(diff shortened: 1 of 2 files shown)


lib/utils/update-check.js
```diff
@@ -25,18 +26,23 @@ module.exports = function updateCheck() {
...
-                    message: 'Continue without upgrading?',
-                    default: false,
-                    name: 'yes'
-                }).then((answers) => {
-                    if (answers.yes) {
-                        return resolve();
-                    }
+                // Yargs hasn't run yet
+                if (includes(process.argv, '--no-prompt')) {
+                    reject('Requested no prompting. Exiting...');
```
acburdine: Might be cleaner just to do:

```js
if (includes(process.argv, '--no-prompt')) {
  reject('Requested no prompting. Exiting...');
  return;
}

inquirer.prompt() // ...
```

ErisDS: This should not reject? Rejecting here means any call to any command with `--no-prompt` will always fail.

rv-aburdine: No - this block of code only gets executed if the version is out of date.

So if your version of ghost-cli is up to date then you'll be fine. If it's out-of-date and you pass the `--no-prompt` flag then it rejects.

ErisDS: See https://github.com/TryGhost/Ghost-CLI/issues/565#issuecomment-350779915. This _only_ makes sense when running the `ghost upgrade` command as far as I can see.

vikaspotluri123: I was thinking of moving the update check to just before yargs runs in [lib/bootstrap](/tryghost/ghost-cli/blob/master/lib/bootstrap.js#L145) based on [#565 (comment)](/tryghost/ghost-cli/issues/565#issuecomment-350769339). It would only resolve / continue if the specific command doesn't require ghost to be up to date (i.e. `CommandClass.versionSpecific = false`)

acburdine:
Closing this as the update-check function now uses the `ui` instance, making this change unnecessary :)

PR #559, opened 2017-12-03
chore(test): implement setup command

vikaspotluri123:
ref: #541


vikaspotluri123:
I thought coveralls drank too much, but it turns out I (figuratively) did 😆

PR #561, opened 2017-12-04
fix(migrate): only run as sudo for sqlite installations

vikaspotluri123:
The migration command needs to run with sudo only if the database is sqlite3 and if the content folder is owned by ghost - otherwise, it can be run as the current user because either the database is mysql, and the credentials are read from the config file, or the content folder is not owned by ghost and we trust the user to have permission to write to it

### DIFF
--- a/lib/tasks/migrate.js
+++ b/lib/tasks/migrate.js
@@ -23,7 +23,7 @@ module.exports = function runMigrations(context) {
 
     // If we're using sqlite and the ghost user owns the content folder, then
     // we should run sudo, otherwise run normally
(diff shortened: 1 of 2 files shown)


kirrg001:
@vikaspotluri123 Was this reported or self-discovered?

vikaspotluri123:
I'm pretty sure it was self-discovered; I don't think I created a related issue.

acburdine:
@vikaspotluri123 Sorry for taking so long to review this! I forgot about it earlier, turns out we _do_ need to run as sudo even for mysql. Knex-migrator outputs log files to the config folder, which requires write access (and therefore sudo).

PR #564, opened 2017-12-05
feat(log): Log improvements for `ghost update` 

### LINKED ISSUE #562: Show help output when `ghost update` fails
If i run `ghost update` and the database migrations fail, i am not getting a hint to run

`ghost update --rollback`

It would be great to show a little hint on the bottom.

CLI: master
Ghost: latest

kirrg001:
closes #562

- be able to pass a command suggestion
- added basic toString fn to error module

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -24,7 +24,6 @@ class CliError extends Error {
         // see https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt
         Error.captureStackTrace(this, this.constructor);
 
-        this.context = options.context || {};
         this.options = options;
 
         this.help = 'Please refer to https://docs.ghost.org/v1/docs/troubleshooting#section-cli-errors for troubleshooting.'
@@ -67,12 +66,30 @@ class CliError extends Error {
     }
 
     /**
+     * Basics to log: message, help and suggestion
+     * @returns {string}
+     */
+    toStringBasics() {
+        let output = `${chalk.yellow('Message:')} ${this.message}\n`;
+
+        if (this.options.help) {
+            output += `${chalk.gray('Help:')} ${this.options.help}\n`;
+        }
+
+        if (this.options.suggestion) {
+            output += `${chalk.green('Suggestion:')} ${this.options.suggestion}\n`;
+        }
+
+        return output;
+    }
+
+    /**
      * Error message string
      * @param {boolean} verbose - Whether or not to render verbose errors
      * @return {string}
      */
     toString(verbose) {
-        let output = `${chalk.yellow('Message:')} ${this.message}\n`;
+        let output = this.toStringBasics();
 
         if (verbose) {
             output += `${chalk.yellow('Stack:')} ${this.stack}\n\n`;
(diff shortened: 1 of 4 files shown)


lib/errors.js
```diff
@@ -22,9 +22,11 @@ class CliError extends Error {
         // see https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt
         Error.captureStackTrace(this, this.constructor);
 
-        this.context = options.context || {};
         this.options = options;
         this.help = 'Please refer to https://docs.ghost.org/v1/docs/troubleshooting#section-cli-errors for troubleshooting.'
+
+        // additional error context
+        this.options.context = this.options.context || {};
```
kirrg001: When i print `this.context` it's never the context object which is passed in the options. It's always a huge object - not sure where this is coming from.

We also use `this.options.help` [here](https://github.com/TryGhost/Ghost-CLI/blob/1.4.0/lib/errors.js#L59)

lib/errors.js
```diff
@@ -51,14 +53,18 @@ class CliError extends Error {
...
-
         if (this.options.help) {
             output += `${chalk.gray('Help:')} ${this.options.help}\n`;
         }
 
+        if (this.options.context.suggestion) {
+            output += `${chalk.green('Suggestion:')} ${this.options.context.suggestion}\n`;
+        }
+
+        if (verbose) {
```
kirrg001: reordered. it looks very unreadable if we first print the stack and then a small line with `Help:`.

kirrg001:
gergregre

kirrg001:
I can add tests if it get's approved 🙃

AileenCGN:
@kirrg001 This PR seems like a good enhancement. Any objections, @acburdine?
Can you add some tests for it?

kirrg001:
> I can add tests if it get's approved 🙃

🙃

kirrg001:
@acburdine Could you pls review? Thanks 👍

lib/errors.js
```diff
@@ -51,14 +53,18 @@ class CliError extends Error {
...
 
-        if (verbose) {
-            output += `${chalk.yellow('Stack:')} ${this.stack}\n`;
-        }
-
         if (this.options.help) {
             output += `${chalk.gray('Help:')} ${this.options.help}\n`;
         }
 
+        if (this.options.context.suggestion) {
```
acburdine: should just make this `this.options.suggestion` and remove the context stuff from above.

lib/errors.js
```diff
@@ -101,8 +114,15 @@ class CliError extends Error {
...
+    constructor(options) {
+        options = options || {};
+
+        options.message = options.message || `Error occurred running command: '${options.cmd}'`;
+        super(options);
+    }
+
     toString(verbose) {
-        let output = chalk.red(`Error occurred running command: '${this.options.cmd}'\n\n`);
+        let output = super.toStringBasics();
```
acburdine: can just use `this.toStringBasics` here b/c the method is inherited.

kirrg001: yeah true 👍

kirrg001:
updated

PR #566, opened 2017-12-12
fix(update-check): only run update check on specific commands

### LINKED ISSUE #565: `ghost -v` breaks programatic upgrade
If I run `ghost -v` I get:

```
You are running an outdated version of Ghost-CLI.
It is recommended that you upgrade before continuing.
Run `npm install -g ghost-cli@latest` to upgrade.

? Continue without upgrading? Yes
```

This forces a prompt.

If I do `ghost -v --no-prompt` it still prompts me.

I am using `ghost -v` to programmatically check if I need to update Ghost or Ghost CLI and I'm now unable to detect that or update blogs, because there's a message in the way that is telling me to update 🙈 .

acburdine:
closes #565, #563
- move update check further into the stack behind a "checkVersion" command flag
- rewrite checkUpdate class to use UI instace
- update tests

### DIFF
--- a/bin/ghost
+++ b/bin/ghost
@@ -4,22 +4,6 @@
 // provide a title to the process
 process.title = 'ghost';
 
-function startup() {
-    const yargs = require('yargs');
-    const bootstrap = require('../lib/bootstrap');
-    bootstrap.run(process.argv.slice(2), yargs);
-}
-
-// Check for CLI updates if not running `ghost run`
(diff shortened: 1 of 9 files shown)


acburdine:
@ErisDS so I enabled version check for a few other commands in which I think it made sense:
- setup command
- install command
- migrate command

All three are commands which I think it is valuable/necessary to have the latest version of Ghost-CLI for.

Issue #230, opened 2017-06-26
Nginx ssl params config is wrong


ErisDS:
See https://github.com/TryGhost/Ghost/issues/8614 for background

in https://github.com/TryGhost/Ghost-CLI/blob/master/lib/services/nginx/files/ssl-params.conf#L12

this line:
> add_header X-Frame-Options DENY;

Prevents ghost exports from working.

I'm not sure where these params came from, but they possibly need review to ensure they're not going to have other side effects.

cc @sebgie

acburdine:
I pulled that config from https://cipherli.st but I found one also at https://mozilla.github.io/server-side-tls/ssl-config-generator/ - the Mozilla one doesn't have the X-Frame-Options header so will look at switching to that config.

sebgie:
No objection to using stricter cypher rules/additional security headers. `X-Frame-Options: SAMEORIGIN` would imo make the download work and still deny malicious attempts to use frames?

PaszaVonPomiot:
@acburdine with nginx config you need to consider compatibility vs security. By not including TLS1.0 you are preventing some Internet Explorer and older Android browsers from openning the page. But that may be a problem if you don't use HSTS Just so you are aware.

I also found that the best cipher setup is following:
`ssl_ciphers AES256+EECDH:AES256+EDH:!aNULL;`
It will have little or none compatibility impact..

kirrg001:
@sebgie Any input on `ssl_ciphers`? Right now we are using `EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH` in the CLI. There are some best practises out there, but i would like to hear your opinion.

jloh:
[Mozillas SSL generator](https://mozilla.github.io/server-side-tls/ssl-config-generator/) suggests these:

```
[...]

PR #568, opened 2017-12-14
removed TLS1

altonius:
TLS1.0 is being deprecated on many websites now (including PCI DSS requiring it for compliance by July 2018).

Is it worth removing this from the ghost TLS/SSL default configuration now?


ErisDS:
Hey @altonius, thanks for the PR. If you're looking to make suggestions like this, they should come with reference materials. You can take a look at https://github.com/TryGhost/Ghost-CLI/issues/230 for an example of the kind of references we use and accept 😊

altonius:
not worries.  I'll retract this for now and will resubmit when we reach July next year when a larger number of sites deprecate TLS1

PR #580, opened 2018-01-07
chore(test): enable macos acceptance tests in travis

acburdine:


acburdine:
Unfortunately macOS builds in travis are pretty slow. Because all of the main developers work using macOS the benefit of running macOS tests each time is fairly minimal, so will close this.

PR #581, opened 2018-01-07

acburdine:
excluded deps:
- validator (because of breaking changes wrt url validation)
- replace-in-file (because of dropped node 4 support)


acburdine:
Odd that updating dependencies would decrease test coverage 🤔

acburdine:
thaaaats weird, with the dependency updates half of command.js becomes uncovered. Wonder if this has something to do with the mocha update.

acburdine:
Update: still investigating which dependency it is that broke test coverage - want to make sure that said dependency didn't break other things as well.

acburdine:
Found out which dependency it was - apparently the rewire major version bump changed some stuff 🤷‍♂️

PR #585, opened 2018-01-11
chore(tests): add uninstall command tests

vikaspotluri123:
refs #541


vikaspotluri123:
And with that, commands are 100% tested 🎉

PR #589, opened 2018-01-18
feat(ui): add `ghost doctor` reference to error messages

AileenCGN:
no issue

Adds information to ui error message that suggests to run `ghost doctor`:

image

@kirrg001 @acburdine Not sure about the wording. For now `ghost doctor` doesn't fix anything, but will check the correct setup, right?


ErisDS:
@AileenCGN correct, however because `ghost doctor` is not a definitive check, I'd adjust the wording slightly: 

> Try running `ghost doctor` to check your system for known issues.

AileenCGN:
@ErisDS Super! Just updated the text!

lib/ui/index.js
```diff
@@ -355,6 +355,7 @@ class UI {
             this.log(debugInfo, 'yellow');
             const logLocation = system.writeErrorLog(stripAnsi(`${debugInfo}\n${output}`));
             this.log(`\nAdditional log info available in: ${logLocation}`);
+            this.log(`\nTry running ${(chalk.cyan('ghost doctor'))} to check your system for known issues.`)
```
kirrg001: Hm shouldn't this be logged for all three cases? I saw lot's of people getting `An unknown error occurred`.

AileenCGN: Yeah. You're right... 🙈 Changes will come in a sec.

acburdine:
So, with the exception of what @kirrg001 noted, I think this change is 👍 . However, something to note is that `ghost doctor`, by default without any args, runs the `install` checks, which aren't going to be super useful for most people debugging an issue _after_ they've installed ghost. Do we want to hold off on adding this to the log output until we have updated `ghost doctor` to be more informative?

kirrg001:
Hmm the install checks might be helpful or?

- nodeVersion
- folderPermissions (!)
- systemStack
- mysql installed

AileenCGN:
@kirrg001 Added the two other cases as well.

@acburdine Yes, I think `ghost doctor` will not fix everything for now, but it's a good start already, and with the coming improvements will be very useful to make to users run, before they open issues. Under the hood, `ghost doctor` will get better - I assume - and lead the user to some potential issues in they're stack.

kirrg001:
> Do we want to hold off on adding this to the log output until we have updated ghost doctor to be more powerful?

What i experienced is, that most people have permission problems. And `ghost doctor` is already able to tell you if your file/folder perms are correct. Not too bad.

I would merge this and push small PR's for `ghost doctor` improvements next week by @AileenCGN (e.g. folder location, logged in user)

And then we can tackle running `ghost doctor` before specific commands.

ErisDS:
Definitely please merge it and ship it!

lib/ui/index.js
```diff
@@ -325,6 +325,7 @@ class UI {
             if (error.logToFile()) {
                 const logLocation = system.writeErrorLog(stripAnsi(`${debugInfo}\n${verboseOutput}`));
                 this.log(`\nAdditional log info available in: ${logLocation}`);
+                this.log(`\nTry running ${(chalk.cyan('ghost doctor'))} to check your system for known issues.`);
```
acburdine: Out of curiosity (sorry I didn't catch this before) - do we want to only output the "ghost doctor" message when the error will log to a file?

For example - any SystemErrors (which Ghost-CLI throws when it runs into certain permissions errors) don't log to a file, so this message won't get output.

AileenCGN: Good point! I'll put it after the conditional, so it'll always be logged!

AileenCGN:
@acburdine @kirrg001 I updated the changes after the last comment and found another place where we can log it additionally. Feel free to merge!

Issue #47, opened 2016-09-30
Ghost Doctor improvements


acburdine:
When speccing out the CLI and its requirements, one of the bigger ones was ensuring that the CLI would have the least chance of failure when installing or updating Ghost. This will hopefully reduce the number of support requests that have to be addressed. 😄 

First, we are establishing the idea of an "official" Ghost stack (see TryGhost/Ghost#7414). However, because there are still oddities that can happen, even with an official environment, more things are needed to ensure the greatest chance of install/update success.

Thus, the `ghost doctor` command exists. This command is designed to run several checks on the system before anything major happens, and the hope is to catch any potential errors _before_ they happen so a more clear error message, as well as a path to fixing/avoiding the errors, can be output.

The `ghost doctor` command also has a concept of "categories", which determine when the specific check is to be run. Essentially, it has one set of checks to be run before installing Ghost for the first time, one set to run before starting ghost, and another set to be run before attempting to update Ghost.

There are already several checks in `ghost doctor` that were copied over from the Ghosts' startup check file. However, there are more that need to be created. They are sorted by category here for convenience
## Install Checks (to be run before `ghost install` does anything)
- ~~check system node version~~ (already completed)
- ~~check file permissions~~ (already completed)
## Startup Checks (to be run before `ghost start`)
- [x] Validate Config (see #26)

## Future improvements!
Although NPM generally does a pretty good job of installing the dependencies that Ghost requires, there are on occasion various glitches that result in different dependencies not being installed correctly. This check would recursively iterate through each dependency and ensure all of the dependencies are installed correctly.

Ghost generally takes ~ 100mb of memory to run well, and just as a sanity check it would probably be a good idea to ensure there is enough free memory on the system to run Ghost. (Node's `os` module has a `freemem()` method that should work for this purpose
## Update Checks (to be run before `ghost update`)
~~- [ ] System has enough memory to run NPM (requires additional research)~~

~~NPM itself can eat up a lot of system memory, and what we _don't_ want to happen is for npm to eat up too much memory and end up crashing an already-running Ghost blog. Similar to the above memory check, the `os.freemem()` method can be used, but additional research is required to see just how much memory npm uses.~~

With the switch to Yarn in #138 - the install process uses significantly less memory, making the necessity of this probably not as great.

## Tasks

See also [comment notes](https://github.com/TryGhost/Ghost-CLI/issues/47#issuecomment-357439800).

### prio
- [x] test current state of the permission checks PR #613 
[...]

PR #591, opened 2018-01-18
feat(startup checks): Ensure enough memory is available before starting

vikaspotluri123:
Refs #47

### DIFF
--- a/lib/commands/doctor/checks/check-memory.js
+++ b/lib/commands/doctor/checks/check-memory.js
@@ -16,5 +16,6 @@ function checkMemory() {
 module.exports = {
     title: 'Checking memory availability',
     task: checkMemory,
+    enabled: ctx => ctx.argv['check-mem'] === true,
     category: ['install', 'start', 'update']
 };
--- a/lib/commands/doctor/checks/check-memory.js
+++ b/lib/commands/doctor/checks/check-memory.js
@@ -1,0 +1,20 @@
+'use strict';
+const os = require('os');
+const SystemError = require('../../../errors').SystemError;
+
+const MB_IN_BYTES = 1048576;
+const MIN_MEMORY = 150;
+
+function checkMemory() {
+    const availableMemory = os.freemem() / MB_IN_BYTES;
+    if (availableMemory < MIN_MEMORY) {
+        return Promise.reject(new SystemError(`Ghost recommends you have at least ${MIN_MEMORY} MB of memory available for smooth operation. It looks like you have ${parseInt(availableMemory)} MB available.`));
+    }
+    return Promise.resolve();
+}
+
+module.exports = {
+    title: 'Checking memory availability',
+    task: checkMemory,
+    category: ['install', 'start', 'update']
+};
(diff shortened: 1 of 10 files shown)


vikaspotluri123:
Should I also add in update checks, or should that be done as a separate PR?

I had started this before winter holidays, but hadn't gotten around to testing it. Just added tests 😄

lib/commands/doctor/index.js
```diff
@@ -18,7 +18,7 @@ class DoctorCommand extends Command {
             return Promise.reject(e);
         }
 
-        return this.ui.listr(checks, {ui: this.ui, system: this.system}).then(() => {
+        return this.ui.listr(checks, {argv: argv, ui: this.ui, system: this.system}).then(() => {
```
acburdine: I'd say we should remove this bit (and the related arguments) for now. Definitely 👍 with being able to skip it - but we can add it back in when we refactor the `ghost doctor` command (which will happen hopefully sometime this week)

vikaspotluri123:
@acburdine 👍 done

acburdine:
@vikaspotluri123 looks like there's some conflicts with the latest changes in master (I refactored the doctor checks & split each check out into a separate file - would you mind doing the same with this change?

vikaspotluri123:
Yeah, I figured it would be better to wait on rebasing until you said something 😜 I'll do it either 2/4 night (CST) or 2/5 morning 👍

acburdine:
No rush - thanks!

vikaspotluri123:
Should I also implement an update mem-check for 200(?) MB?

acburdine:
Hmm....I'd say maybe just increase the memory limit to 150mb and run it on both start and update? That should be enough I think.

vikaspotluri123:
Not sure if I explicitly said it, but this is ready 👍

lib/commands/doctor/index.js
vikaspotluri123: With this am I good to re-add the skip?

acburdine: Sorry for not responding to this until now - I'm good with re-adding the skip. However, that does mean that any command that _calls_ ghost doctor (e.g. `start`, `install`, `update`) will need to be updated to also pull in the required `ghost doctor` args.

vikaspotluri123: Done! 👍

lib/commands/doctor/checks/check-memory.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+const os = require('os');
+const SystemError = require('../../../errors').SystemError;
+
+const MB_IN_BYTES = 1048576;
+const MIN_MEMORY = 150;
+
+function checkMemory(ctx, task) {
+    if (!ctx.argv['check-mem']) {
+        return task.skip();
```
acburdine: I wonder if it's worth moving this to an `enable` function? That way the check wouldn't show up in the list if it's been intentionally ignored 🤔

vikaspotluri123: Done!

AileenCGN:
Hey @acburdine is there a reason why this is approved but not merged yet?

vikaspotluri123:
Isn't the option already there?

acburdine:
yeah it's there - what I meant by "push" is advertise it more in the docs & such.

Issue #587, opened 2018-01-16
Optimise error messages


kirrg001:
This issue lists all the places where we have to optimise the error message.

- [x] https://github.com/TryGhost/Ghost-CLI/blob/1.4.2/extensions/nginx/acme.js#L52 - if i am running into an error here, the error properties from the original error are not taken into account and it shows me a too generic message (```{ Error: EACCES: permission denied, rmdir '/tmp/acme.sh/Neilpang-acme.sh-b6c2fc5/.github'
    at Error (native)
  errno: -13,
  code: 'EACCES',
  syscall: 'rmdir',
  path: '/tmp/acme.sh/Neilpang-acme.sh-b6c2fc5/.github' }``` - you can use this as mock (!). This is just one example case. The error handler should be able to take custom properties into account!

------

Edit Aileen:

Tasks:
- [ ] error handling extensions (missing catches, better error messages)
- [ ] error handling core (missing catches, better error messages)
- [ ] improve `ProcessError` so it can deal with custom properties

PR #595, opened 2018-01-24
feat(ssl) improve error messages to log original error

AileenCGN:
refs #587

- fixes an issue, where failed requests with `got` wouldn't get catched properly
- differentiates between `ProcessError` for `stderr` and `CliError` for everything else
- `CliError` can be passed the original error and will log the `message` and `stack` to the log file, or log in the UI when in verbose mode
- Adds `err` property with original error whereever we have it so it'll be passed along and printed in the log file

### DIFF
--- a/extensions/nginx/acme.js
+++ b/extensions/nginx/acme.js
@@ -27,14 +27,13 @@ function install(ui, task) {
         ui.logVerbose('ssl: downloading acme.sh to temporary directory', 'green');
         return fs.emptyDir(acmeTmpDir)
     }).then(() => got(acmeApiUrl)).then((response) => {
-        if (response.statusCode !== 200) {
-            return Promise.reject(new cli.errors.CliError('Unable to query GitHub for ACME download URL'));
-        }
-
         try {
             response = JSON.parse(response.body).tarball_url;
         } catch (e) {
-            return Promise.reject(new cli.errors.CliError('Unable to parse Github api response for acme'));
+            return Promise.reject(new cli.errors.CliError({
+                message: 'Unable to parse Github api response for acme',
+                err: e
+            }));
         }
 
         return download(response, acmeTmpDir, {extract: true});
@@ -49,7 +48,21 @@ function install(ui, task) {
 
         // Installs acme.sh into /etc/letsencrypt
         return ui.sudo('./acme.sh --install --home /etc/letsencrypt', {cwd: acmeCodeDir});
-    }).catch((error) => Promise.reject(new cli.errors.ProcessError(error)));
+    }).catch((error) => {
+        // CASE: error is already a cli error, just pass it along
+        if (error instanceof cli.errors.CliError) {
+            return Promise.reject(error);
+        }
+
+        // catch any request errors first, which isn't a ProcessError
+        if (!error.stderr) {
+            return Promise.reject(new cli.errors.CliError({
+                message: 'Unable to query GitHub for ACME download URL',
+                err: error
+            }));
+        }
+        return Promise.reject(new cli.errors.ProcessError(error));
+    });
 }
 
 function generateCert(ui, domain, webroot, email, staging) {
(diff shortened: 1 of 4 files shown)


extensions/nginx/acme.js
```diff
@@ -64,13 +74,19 @@ function generateCert(ui, domain, webroot, email, staging) {
 
         if (error.stderr.match(/Verify error:(Fetching|Invalid Response)/)) {
             // Domain verification failed
-            return Promise.reject(new cli.errors.SystemError(
-                'Your domain name is not pointing to the correct IP address of your server, please update it and run `ghost setup ssl` again'
-            ));
+            return Promise.reject(new cli.errors.SystemError({
```
AileenCGN: @acburdine I don't understand why we're using `SystemError` here, but further above `CliError`. What is the intention?

extensions/nginx/acme.js
```diff
@@ -27,10 +27,6 @@ function install(ui, task) {
         ui.logVerbose('ssl: downloading acme.sh to temporary directory', 'green');
         return fs.emptyDir(acmeTmpDir)
     }).then(() => got(acmeApiUrl)).then((response) => {
-        if (response.statusCode !== 200) {
```
AileenCGN: I removed this part from here, as `got` should never resolve if the `statusCode` is not `2xx`. See: https://github.com/sindresorhus/got#gothttperror

Unless you can think of any other occurrence?

lib/errors.js
```diff
@@ -23,7 +23,10 @@ class CliError extends Error {
         Error.captureStackTrace(this, this.constructor);
 
         this.context = options.context || {};
+        // TODO: Do something with the original error message which we can receive now as the `err` property
+        this.err = options.err || {};
```
AileenCGN: @kirrg001 What is your opinion on going further here? Does it make more sense to pick the possible error properties where they occur? For example, when we know the error [here](https://github.com/TryGhost/Ghost-CLI/pull/595/files#diff-f158eddc93a4bd540b26f3dc0da0527dR55) should almost always be a request error, shall we rely on that and pick the properties that `got` returns? Or just pass the full original error and log it into the log files?

kirrg001: I would try to keep this as simple as possible.

e.g. log custom CLI error, log original error if exists (e.g. iterate over the properties and log each field). Maybe we can only log the original error into the debug file.

custom.message
custom.stack
original.message
original.stack

lib/errors.js
```diff
@@ -83,11 +114,16 @@ class ProcessError extends CliError {
             output += chalk.yellow(`Exit code: ${this.options.code}\n\n`);
         }
 
-        if (verbose && this.options.stdout) {
```
AileenCGN: When testing different error cases, I stumbled over a situation where `stderr` was filled, but `stdout` was empty. I changed the logic here, so it doesn't depend on `stdout` only to log our error.

extensions/nginx/acme.js
vikaspotluri123: I'm pretty sure that was me being a goof 😜

AileenCGN:
@kirrg001 @acburdine I think this one is ready for review!

lib/errors.js
```diff
@@ -10,6 +10,8 @@ const chalk = require('chalk');
  */
 class CliError extends Error {
     constructor(options) {
+        const originalError = {};
```
kirrg001: This is not a const.

AileenCGN: Hmmm... weird... Because `lint` tells me different:

```
yarn run v1.3.2
$ eslint bin lib test extensions

/Users/aileen/code/Ghost-CLI/lib/errors.js
  13:13  error  'originalError' is never reassigned. Use 'const' instead  prefer-const

✖ 1 problem (1 error, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.

error Command failed with exit code 1.
```

acburdine: this would be a const, yes. Because it's being assigned to an object, you can mutate that object's properties regardless of whether or not it's a const.

kirrg001: I didn't knew THAT 😁

kirrg001:
@acburdine Thanks for code review. I'll test that PR later 👍

lib/utils/resolve-version.js
```diff
@@ -53,7 +53,8 @@ module.exports = function resolveVersion(version, update) {
         } catch (e) {
             return Promise.reject(new errors.CliError({
                 message: 'Ghost-CLI was unable to load versions from Yarn.',
-                log: false
+                log: false,
+                err: e
```
kirrg001: `log` is false. So we don't log into file.

And we only log the original error in verbose mode. That means I won't see the original error?

PR #597, opened 2018-01-25
feat(doctor) prevent root user from installing

AileenCGN:
refs #47

- adds a `currentUser` check, which detects if the user tries to install ghost as `root` and prevents if so
- the task will be skipped for a local install
- adds tests

### DIFF
--- a/lib/commands/doctor/checks/install.js
+++ b/lib/commands/doctor/checks/install.js
@@ -68,6 +68,19 @@ const tasks = {
             return tasks.checkDirectoryAndAbove(process.cwd(), 'run `ghost install`');
         });
     },
+    currentUser: function currentUser() {
+        // Check if user is trying to install as `root`
+        if (process.getuid() === 0) {
+            return Promise.reject(new errors.SystemError(
+                `Can't install as 'root' user.${eol}` +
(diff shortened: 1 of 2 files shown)


lib/commands/doctor/checks/install.js
```diff
@@ -68,6 +68,19 @@ const tasks = {
             return tasks.checkDirectoryAndAbove(process.cwd(), 'run `ghost install`');
         });
     },
+    currentUser: function currentUser() {
+        // Check if user is trying to install as `root`
+        if (process.getuid() === 0) {
+            return Promise.reject(new errors.SystemError(
+                `Can't install as 'root' user.${eol}` +
+                `Please create a new user with regular account privileges to install Ghost.${eol}` +
+                `See ${chalk.underline.blue('https://docs.ghost.org/docs/install')} ` +
```
acburdine: Do we want to link to the specific section of the page that talks about the non-root user or just to the page itself?

AileenCGN: Yeah, was thinking to do that but than thought that's it might be to volatile. But you're right... I'll change it to link to the section.

kirrg001:
Will also test this PR later!

AileenCGN:
Closing this in favour of #604

PR #600, opened 2018-01-30
[wip] use systemd-notify for success and error calls

acburdine:
[ci skip]
- update ghost service template
- add success/error handlers to systemd process manager

TODO:
- [ ] test this _heavily_ to make sure it works
- [ ] add migration if it does actually work
- [ ] add unit tests


acburdine:
Closing because of #612

Issue #540, opened 2017-11-16
Investigate using systeminformation to have a more unified/reliable way of determining system information


acburdine:
https://github.com/sebhildebrandt/systeminformation -> could replace some of our various shell calls for system data.

acburdine:
We're using it for memory availability now, so I'm going to close this. If we find other places that we can refactor to use systeminformation in the future we can handle that on a case-by-case basis 😄

PR #602, opened 2018-01-30
feat(system) add os and version to System class

AileenCGN:
no issue

- added a new getter to the System class, that returns the operating system as well as the version
- works for linux, mac, and windows. Everything else falls back to the native `os.platform()` fn.
- prints the os and the version in the debug information
- added tests

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -40,7 +40,7 @@ Paste log file here
 
 This is automatically output by Ghost-CLI if an error occurs, please copy & paste:
 
-* OS (add version):
+* OS:
 * Node Version:
 * Ghost-CLI Version:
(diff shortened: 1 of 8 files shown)


acburdine:
Out of curiosity, is there a reason we're using our own commands and not something like https://www.npmjs.com/package/systeminformation ? imo using a library for it is probably a better approach here.

AileenCGN:
Thanks for the info. I will look into it, but with using `systeminformation` we'd need to refactor the code to be async.

If we decide to keep our own commands, I will source the fn out into a util, but still keep the command accessible from our System class.

acburdine:
Sounds good to me - any refactoring can be done as part of addressing #540

AileenCGN:
@acburdine This is ready for review!

PR #603, opened 2018-01-30
fix(nginx tests): stub OS for migration skip tests

vikaspotluri123:
refs #541  - see [comment](https://github.com/TryGhost/Ghost-CLI/issues/541#issuecomment-361513383)

cc @AileenCGN - can you check that this fixes the problem? I'm 95% sure it will 😄


vikaspotluri123:
Build 1234 😆

PR #604, opened 2018-02-01
feat(command) prevent commands run as root user

AileenCGN:
refs #47

- adds a `checkRootUser` fn to our base command class, which prevents any command to be run as `root`
- adds test

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -167,6 +167,12 @@ class Command {
             checkValidInstall(commandName);
         }
 
+        if (!this.allowRoot) {
+            const checkRootUser = require('./utils/check-root-user');
+            // Check if user is trying to install as `root`
+            checkRootUser();
+        }
+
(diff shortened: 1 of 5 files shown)


alikuru:
It would be great to have a switch that re-enable running as root.

ErisDS:
Running as root has never been supported - our docs have always made this clear, however we have run into numerous issues with this advice being ignored and subsequently causing support & maintenance overhead for the project, including the DO one-click installer incorrectly using root.

Our goal is to keep Ghost-CLI maintainable for our small team ([read more here](https://github.com/TryGhost/Ghost-CLI#project-goals). Therefore we are actively working to get users off of setups that cause support and maintenance overhead.

I saw on slack that you're wanting to run using pm2 - this will be supported using extensions. We will not be adding a flag for running as root.

PR #607, opened 2018-02-02
fix(instance): query process managers in dev & prod if running not set

### LINKED ISSUE #463: Make `ghost run` be responsible for outputting running environment

acburdine:
closes #463
- query each process manager if running is not set in cli config

### DIFF
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -111,6 +111,30 @@ class Instance {
         }
 
         if (!this.cliConfig.has('running')) {
+            const currentEnvironment = this.system.development;
+
+            // Check production environment first
+            if (Config.exists(path.join(this.dir, 'config.production.json'))) {
+                this.system.setEnvironment(false);
+
+                if (this.process.isRunning(this.dir)) {
+                    this.cliConfig.set('running', 'production').save();
+                    return true;
+                }
+            }
+
+            // Check development next
+            if (Config.exists(path.join(this.dir, 'config.development.json'))) {
+                this.system.setEnvironment(true);
+
+                if (this.process.isRunning(this.dir)) {
+                    this.cliConfig.set('running', 'development').save();
+                    return true;
+                }
+            }
+
+            this.system.setEnvironment(currentEnvironment);
+
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


kirrg001:
Just leaving this here.

---

You have a dev and prod config file. Both use the `local` process manager. You start your development blog (`ghost start`). Then you loose the information that your instance is running in your `.ghost-cli` file. You do `ghost ls` and it would show that production is running. This could be solved by writing the PID including the env information?


---

You install a production blog on linux. Then you run `cp config.production.json config.development.json`. No modifications. You stop your blog. Your run `ghost start` and you will see `running (development)`. Not sure how that is possible.

PR #610, opened 2018-02-02
feat(doctor) content folder permission checks

AileenCGN:
refs #47

- adds a new task to `ghost doctor startup` which runs on `ghost start`
- skips this task if
	- ghost user is not setup or
	- ghost user doesn't own the content folder
- checks the permissions inside the content folder and fails if a file or directory is not owned by ghost
- logs the relevant files or directories
- adds tests

### DIFF
--- a/lib/commands/doctor/checks/content-folder.js
+++ b/lib/commands/doctor/checks/content-folder.js
@@ -1,0 +1,43 @@
+'use strict';
+const execa = require('execa');
+const path = require('path');
+const Promise = require('bluebird');
+const chalk = require('chalk');
+
+const errors = require('../../../errors');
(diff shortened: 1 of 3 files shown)


lib/commands/doctor/checks/content-folder.js
```diff
@@ -0,0 +1,45 @@
...
+        if (error instanceof errors.SystemError) {
+            return Promise.reject(error);
+        }
+        return Promise.reject(new errors.ProcessError(error));
+    });
+}
+
+module.exports = {
+    title: 'Content folder permissions',
+    enabled: () => {
```
acburdine: this can be just `() => shouldUseGhostUser...`

PR #611, opened 2018-02-03
fix(setup): skip ghost user setup if process manager is local

acburdine:
no issue
- noticed this when testing #607

### DIFF
--- a/lib/commands/setup.js
+++ b/lib/commands/setup.js
@@ -111,7 +111,7 @@ class SetupCommand extends Command {
             }
         }];
 
-        if (!argv.local && argv['setup-linux-user'] !== false && os.platform() === 'linux') {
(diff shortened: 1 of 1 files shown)


lib/commands/setup.js
```diff
@@ -111,7 +111,7 @@ class SetupCommand extends Command {
             }
         }];
 
-        if (!argv.local && argv['setup-linux-user'] !== false && os.platform() === 'linux') {
+        if (!argv.local && argv['setup-linux-user'] !== false && os.platform() === 'linux' && argv.process !== 'local') {
```
vikaspotluri123: If `argv.process == local`, do we need to check `argv.local`?

acburdine: probably not

acburdine: will leave it in there though - doesn't really hurt anything and it's good just to make sure 😄

PR #612, opened 2018-02-03
feat(core): Ensure that Ghost was started

### LINKED ISSUE #472: Implement better ipc communication to ghost run process
One thing that can be somewhat confusing currently when using the systemd process manager in the CLI is the fact that `ghost start` exits quickly with the message that ghost has started, when in fact it can sometimes take 20-30 seconds to actually start (time may vary from server to server based on a number of factors). Conversely, when you run `ghost start` using the local process manager, it doesn't exit as quickly, but when it _does_ exit, ghost has started.

The reason for this difference is that the local process manager spawns the `ghost run` subprocess directly, and thus can communicate with it over an IPC tunnel, exiting once ghost has fully started. Systemd, on the other hand, must go through the intermediary process of `systemctl`, which has no way of letting the ghost start process know once the process has completely started. Other process manager extensions may also have the same issue depending on how they're implemented and how the underlying process manager functions.

To make things more consistent between process managers, the CLI should use a more advanced method of ipc tunneling (using something like https://www.npmjs.com/package/node-ipc).

Note: this would also potentially open up opportunities for improvement in other areas: for example, `ghost update` could interface with the ghost run process and tell it to run in maintenance mode before modifying the database.

kirrg001:
closes #472

- added port polling utility
- general process manager class offers `ensureStarted` function
- systemd extension makes use of `ensureStarted`

### DIFF
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -1,6 +1,5 @@
 'use strict';
 const spawn = require('child_process').spawn;
-const merge = require('lodash/merge');
 const Command = require('../command');
 
 class RunCommand extends Command {
@@ -55,9 +54,7 @@ class RunCommand extends Command {
         this.sudo = true;
     }
 
-    useDirect(instance, options) {
-        options = merge({delayErrorChaining: true}, options || {});
-
+    useDirect(instance) {
         this.child = spawn(process.execPath, ['current/index.js'], {
             cwd: instance.dir,
             stdio: [0, 1, 2, 'ipc']
@@ -74,12 +71,8 @@ class RunCommand extends Command {
                 return;
             }
 
-            if (!options.delayErrorChaining) {
-                instance.process.error(message.error);
-            } else {
-                // NOTE: Backwards compatibility of https://github.com/TryGhost/Ghost/pull/9440
-                setTimeout(() => {instance.process.error(message.error);}, 1000);
-            }
+            // NOTE: Backwards compatibility of https://github.com/TryGhost/Ghost/pull/9440
+            setTimeout(() => {instance.process.error(message.error);}, 1000);
         });
     }
 
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -1,5 +1,6 @@
 'use strict';
 const spawn = require('child_process').spawn;
+const merge = require('lodash/merge');
 const Command = require('../command');
 
 class RunCommand extends Command {
@@ -54,7 +55,9 @@ class RunCommand extends Command {
         this.sudo = true;
     }
 
-    useDirect(instance) {
+    useDirect(instance, options) {
+        options = merge({delayErrorChaining: true}, options || {});
+
         this.child = spawn(process.execPath, ['current/index.js'], {
             cwd: instance.dir,
             stdio: [0, 1, 2, 'ipc']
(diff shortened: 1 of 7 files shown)


lib/commands/run.js
```diff
@@ -71,7 +71,11 @@ class RunCommand extends Command {
                 return;
             }
 
-            instance.process.error(message.error);
+            // Ghost error log happens AFTER this event
+            // REMOVE WHEN ghost is released and we bump the cli engine?
+            setTimeout(() => {
```
kirrg001: See https://github.com/TryGhost/Ghost/blob/1.20.3/index.js#L44.
Logging happens to late. PR to Ghost required.
For backwards support, we should keep a timeout here?

lib/commands/start.js
```diff
@@ -37,10 +37,16 @@ class StartCommand extends Command {
...
+                    .then(() => {
+                        return processInstance.ensureStarted(process.cwd(), this.system.environment);
+                    })
+                    .then(() => {
+                        instance.running(this.system.environment);
+                    });
             };
 
             return this.ui.run(start, 'Starting Ghost', runOptions).then(() => {
+                // @TODO: shouldn't this happen before start?
```
kirrg001: Needs a little discussion. I wondered why we enable the target process after we've started the instance.
If there is a good reason, we should explain it here.

lib/process-manager.js
```diff
@@ -60,6 +66,68 @@ class ProcessManager {
         // Base Implementation
     }
 
+    /**
+     * General implementation of figuring out if the Ghost blog has started successfully.
```
kirrg001: We can reconsider if we want the port polling implementation only in the systemd extension.
Local process manger overrides it currently. I am fine with both. My thinking is that other extensions will find that helpful for sure.

lib/commands/run.js
kirrg001: [Ghost PR](https://github.com/TryGhost/Ghost/pull/9440)

lib/process-manager.js
```diff
@@ -60,6 +63,33 @@ class ProcessManager {
...
+     *
+     * @returns {Promise<any>}
+     */
+    ensureStarted(options) {
+        options = merge({
```
acburdine: should be replaced with Object.assign

lib/utils/port-polling.js
```diff
@@ -0,0 +1,104 @@
+'use strict';
+
+const net = require('net');
+const merge = require('lodash/merge');
+const errors = require('../errors');
+
+module.exports = function portPolling(options) {
+    options = merge({
```
acburdine: Should be replaced with Object.assign

lib/process-manager.js
```diff
@@ -60,6 +63,33 @@ class ProcessManager {
         // Base Implementation
     }
 
+    /**
+     * General implementation of figuring out if the Ghost blog has started successfully.
+     *
+     * @returns {Promise<any>}
+     */
+    ensureStarted(options) {
```
acburdine: we should require the port polling utility inside this function, so `ghost run` doesn't have to require it.

kirrg001: 👍

acburdine:
tested this and it works 👍 just need the above changes fixed and this is good to go.

kirrg001:
@acburdine cool 👍 will push your suggestions

lib/process-manager.js
kirrg001: 👍

kirrg001:
Pushed 2 commits 👍

PR #613, opened 2018-02-03
feat(doctor) detect incorrect permissions

### LINKED ISSUE #294: Ghost folder & file permissions
In an ideal world, we'd set the folder permissions to be 774 and the file permissions to be 664 for everything inside a ghost directory:

e.g.

`find /path -type d -exec chmod 774 {} \;`
`find /path -type f -exec chmod 664 {} \;`

For now they have been set to 774 as a shortcut to a good, if not the best, solution.

This would need to be done on initial install, and probably rechecked/rerun on version folders after an install.

AileenCGN:
refs #47
closes #294

- Renames `folderPermissions` to `installFolderPermissions`
- Adds two more tasks: `filePermissions` and `folderPermissions` which check for the correct permissions inside of the ghost installation folder
	- not enabled when local processmanager is used (= local install)
	- calls `checkPermissions` util
	- `filePermissions` does not check the versions and current folder
- Refactors the `contentFolderPermissions` to be named `contentFolder` and disables task when content folder is not owned by ghost or ghost user is not found.
- Adds `checkPermissions` util which takes a property to specify which check should run:
	- `owner` checks for ownership of the content folder
	- `folder` checks if directories within the ghost installation folder have the correct permission (775)
	- `files` checks if files within the ghost installation folder have the correct permissions (664)
- Adds tests

### DIFF
--- a/lib/commands/doctor/checks/check-permissions.js
+++ b/lib/commands/doctor/checks/check-permissions.js
@@ -1,0 +1,62 @@
+'use strict';
+
+const execa = require('execa');
+const chalk = require('chalk');
+
+const errors = require('../../../errors');
+
+module.exports = function checkPermissions(type) {
+    let errMsg;
+
+    // fall back to check the owner permission of the content folder if nothing specified
+    type = type || 'owner';
+
+    const checkTypes = {
+        owner: {
+            command: 'find ./content ! -group ghost ! -user ghost',
+            help: `Run ${chalk.green('sudo chown -R ghost:ghost ./content')} and try again.`
+        },
+        folder: {
+            command: 'find ./ -type d ! -perm 775 ! -perm 755',
+            help: `Run ${chalk.green('sudo find ./ -type d -exec chmod 775 {} \\;')} and try again.`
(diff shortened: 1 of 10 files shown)


test/unit/commands/doctor/startup-spec.js
```diff
@@ -8,157 +8,342 @@ const setupEnv = require('../../../utils/env');
...
+            return configValidationCheck[0].task({environment: 'testing'}).then(() => {
+                env.cleanup();
+                expect(false, 'error should have been thrown').to.be.true;
+            }).catch((error) => {
+                expect(error).to.be.an.instanceof(errors.ConfigError);
+                expect(error.message).to.match(/Config file is not valid JSON/);
+                expect(error.options.environment).to.equal('testing');
+                expect(cwdStub.calledOnce).to.be.true;
+
+                env.cleanup();
```
vikaspotluri123: This isn't related specifically to the PR, but we should probably move this to the top of the block - if any of the assertions fail, this will never be called

PR #614, opened 2018-02-03
Ghost doctor refactor

acburdine:
TODO:
- [x] split doctor checks out into separate files
- [x] refactor doctor command
- [x] fix usage of doctor command in other commands
~~[ ] add extension hook ability~~ can do later


lib/commands/doctor/checks/mysql.js
```diff
@@ -0,0 +1,43 @@
...
+    // So, we modify the path env var to make things work
+    const options = ctx.system.platform.linux ? {env: {PATH: `/usr/sbin:${process.env.PATH}`}} : {};
+
+    // Technically this doesn't work on windows, but there's
+    // not an easy way to do that anyways so ¯\_(ツ)_/¯
+    return execa.shell('which mysqld', options).catch(() => {
+        ctx.ui.log(`${chalk.yellow(`Local MySQL install not found. You can ignore this if you are using a remote MySQL host.
+Alternatively you could:`)}
+${chalk.blue('a)')} install MySQL locally
+${chalk.blue('b)')} run ${chalk.cyan('`ghost install --db=sqlite3`')} to use sqlite
```
vikaspotluri123: This clashes with the [docs](https://docs.ghost.org/docs/cli-install) wrt the `dbpath` flag

acburdine: The docs are incorrect (I think) there - you don't have to pass the dbpath flag.

PR #618, opened 2018-02-04
feat: Run all doctor checks & print summary

acburdine:


lib/commands/doctor/checks/system-stack.js
```diff
@@ -35,24 +37,35 @@ function systemStack(ctx) {
...
+            Promise.resolve({yes: false})
+        ).then((answer) => {
+            if (answer.yes) {
+                task.skip('System stack check skipped');
+                return Promise.resolve();
+            }
+
+            return Promise.reject(new errors.SystemError({
+                message: `System stack checks failed with message: '${error.message}'`,
+                task: 'Checking operating system compatibility'
```
kirrg001: I guess you can use `taskTitle` here.

acburdine: oops 🙈

lib/commands/doctor/checks/system-stack.js
```diff
@@ -35,24 +37,35 @@ function systemStack(ctx) {
...
+    });
 }
 
 module.exports = {
-    title: 'Checking operating system compatibility',
+    title: taskTitle,
     task: systemStack,
     enabled: (ctx) => !ctx.local,
-    skip: (ctx) => ctx.argv && !ctx.argv.stack,
+    skip: (ctx) => !ctx.isDoctorCommand && ctx.argv && !ctx.argv.stack,
```
kirrg001: What is the diff between `ctx.local` and `argv.stack`? 🤔 
Aren't they the same somehow? `argv.stack` is false if local PM is used.

kirrg001: Discussed.

`--no-stack` => `argv.stack=false` (but can be local PM or a different extension)

PR #623, opened 2018-02-06
fix(doctor) skip validate config check for running instances

AileenCGN:
no issue

- skips the validate config check when the tested instance is currently running, as it would fail for testing the server port

### DIFF
--- a/lib/commands/doctor/checks/validate-config.js
+++ b/lib/commands/doctor/checks/validate-config.js
@@ -49,5 +49,6 @@ function validateConfig(ctx) {
 module.exports = {
     title: taskTitle,
     task: validateConfig,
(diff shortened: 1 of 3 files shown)


lib/commands/doctor/checks/validate-config.js
```diff
@@ -49,5 +49,10 @@ function validateConfig(ctx) {
 module.exports = {
     title: taskTitle,
     task: validateConfig,
+    skip: (ctx) => {
+        const instance = ctx.system.getInstance();
```
acburdine: I'd say this could be a lot shorter:

```js
skip: (ctx) => !ctx.instance || ctx.instance.process.isRunning(ctx.instance.dir);
```

AileenCGN: 👍

PR #631, opened 2018-02-06
fix(root-user): show different error message for 1-click installs

AileenCGN:
no issue

- detect DigitalOcean One-Click install and render a message how to migrate their installation to a non-root user

### DIFF
--- a/lib/utils/check-root-user.js
+++ b/lib/utils/check-root-user.js
@@ -2,18 +2,41 @@
 
 const os = require('os');
 const chalk = require('chalk');
+const fs = require('fs');
+const includes = require('lodash/includes');
 
-function checkRootUser() {
-    // Skip if we're on windows
-    if (os.platform() !== 'linux') {
+const isRootInstall = function isRootInstall() {
+    const path = require('path');
+    const cliFile = path.join(process.cwd(), '.ghost-cli');
+
+    return fs.existsSync(cliFile) && fs.statSync(cliFile).uid === 0;
+}
+
+function checkRootUser(command) {
+    const allowedCommands = ['stop', 'start', 'restart'];
+    const isOneClickInstall = fs.existsSync('/root/.digitalocean_password');
+
+    if (os.platform() !== 'linux' || process.getuid() !== 0) {
         return;
     }
 
-    if (process.getuid() === 0) {
+    if (isOneClickInstall) {
+        // We have a Digitalocean one click installation
+        console.error(`${chalk.yellow('We discovered that you are using the Digitalocean One-Click install.')}
+You need to create a user with regular account privileges and migrate your installation to use this user.
+Please follow the steps here: ${chalk.underline.green('https://docs.ghost.org/docs/troubleshooting#section-fix-root-user')} to fix your setup.\n`);
+    } else if (isRootInstall()) {
+        console.error(`${chalk.yellow('It seems Ghost was installed using the root user.')}
+You need to create a user with regular account privileges and migrate your installation to use this user.
+Please follow the steps here: ${chalk.underline.green('https://docs.ghost.org/docs/troubleshooting#section-fix-root-user')} to fix your setup.\n`);
+    } else {
         console.error(`${chalk.yellow('Can\'t run command as \'root\' user.')}
-Please create a new user with regular account privileges and use this user to run the command.
-See ${chalk.underline.blue('https://docs.ghost.org/docs/install#section-create-a-new-user')} for more information`);
+Please use the user you set up in the installation process, or create a new user with regular account privileges and use this user to run 'ghost ${command}'.
+See ${chalk.underline.green('https://docs.ghost.org/docs/install#section-create-a-new-user')} for more information\n`);
+    }
 
+    // TODO: remove this 4 versions after 1.5.0
+    if (!includes(allowedCommands, command)) {
         process.exit(1);
     }
 }
--- a/lib/command.js
+++ b/lib/command.js
@@ -117,7 +117,7 @@ class Command {
         if (!this.allowRoot) {
             const checkRootUser = require('./utils/check-root-user');
             // Check if user is trying to install as `root`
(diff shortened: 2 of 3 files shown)


lib/utils/check-root-user.js
```diff
@@ -2,20 +2,26 @@
...
+    if (os.platform() !== 'linux' || process.getuid() !== 0) {
         return;
     }
 
-    if (process.getuid() === 0) {
+    if (fs.existsSync('/root/.digitalocean_password')) {
+        // We have a Digitalocean one click installation
+        console.error(`${chalk.yellow('We discovered that you are using the Digitalocean 1-click install.')}
+You need to create a user with regular account privileges and migrate your installation to use this user.
+Please follow the steps here ${chalk.underline.green('https://docs.ghost.org/docs/troubleshooting#section-fix-root-user')} to fix your setup.`);
```
kirrg001: When i run `find . -group root -user root -exec chown <user>:<user> {} \;
` on my one-click DO, my `current folder` is still owned by root?


> lrwxrwxrwx 1 root  root    29 Jan 24 16:59 current -> /var/www/ghost/versions/1.6.2/

kirrg001: Probably because it's a symlink.

acburdine: yeah - I think that's fine - although you might test running `ghost update` on a DO 1-click install to see if it's able to recreate the symlink after fixing the install.

kirrg001: @AileenCGN can you test that pls? i destroyed my droplet already

lib/utils/check-root-user.js
```diff
@@ -2,20 +2,53 @@
...
-See ${chalk.underline.blue('https://docs.ghost.org/docs/install#section-create-a-new-user')} for more information`);
+Please use the user you set up in the installation process, or create a new user with regular account privileges and use this user to run 'ghost ${command}'.
+See ${chalk.underline.green('https://docs.ghost.org/docs/install#section-create-a-new-user')} for more information`);
+    }
 
+    if (!includes(allowedCommands, command)) {
         process.exit(1);
     }
+
+    return;
```
acburdine: we don't need this return here

lib/utils/check-root-user.js
```diff
@@ -2,20 +2,53 @@
...
+Please follow the steps here ${chalk.underline.green('https://docs.ghost.org/docs/troubleshooting#section-fix-root-user')} to fix your setup.`);
+    } else {
         console.error(`${chalk.yellow('Can\'t run command as \'root\' user.')}
-Please create a new user with regular account privileges and use this user to run the command.
-See ${chalk.underline.blue('https://docs.ghost.org/docs/install#section-create-a-new-user')} for more information`);
+Please use the user you set up in the installation process, or create a new user with regular account privileges and use this user to run 'ghost ${command}'.
+See ${chalk.underline.green('https://docs.ghost.org/docs/install#section-create-a-new-user')} for more information`);
+    }
 
+    if (!includes(allowedCommands, command)) {
```
acburdine: we talked about last night removing the "allowed commands" 3 or 4 versions from now - can you add a comment to this here saying as much?

lib/utils/check-root-user.js
```diff
@@ -2,20 +2,53 @@
...
     }
 
-    if (process.getuid() === 0) {
+    if (isOneClickInstall) {
+        // We have a Digitalocean one click installation
+        console.error(`${chalk.yellow('We discovered that you are using the Digitalocean One-Click install.')}
+    You need to create a user with regular account privileges and migrate your installation to use this user.
+    Please follow the steps here ${chalk.underline.green('https://docs.ghost.org/docs/troubleshooting#section-fix-root-user')} to fix your setup.`);
+    } else if (isRootInstall()) {
+        console.error(`${chalk.yellow('It seems installed Ghost as root user.')}
```
acburdine: Should be "It seems Ghost was installed using the root user"

lib/utils/check-root-user.js
```diff
@@ -2,20 +2,53 @@
...
+    if (os.platform() !== 'linux' || process.getuid() !== 0) {
         return;
     }
 
-    if (process.getuid() === 0) {
+    if (isOneClickInstall) {
+        // We have a Digitalocean one click installation
+        console.error(`${chalk.yellow('We discovered that you are using the Digitalocean One-Click install.')}
+    You need to create a user with regular account privileges and migrate your installation to use this user.
+    Please follow the steps here ${chalk.underline.green('https://docs.ghost.org/docs/troubleshooting#section-fix-root-user')} to fix your setup.`);
```
acburdine: maybe add a colon after "here"? `Please follow the steps here: ${...`

lib/utils/check-root-user.js
```diff
@@ -2,20 +2,53 @@
...
+    } else if (isRootInstall()) {
+        console.error(`${chalk.yellow('It seems installed Ghost as root user.')}
+You need to create a user with regular account privileges and migrate your installation to use this user.
+Please follow the steps here ${chalk.underline.green('https://docs.ghost.org/docs/troubleshooting#section-fix-root-user')} to fix your setup.`);
```
acburdine: same thing with the colon here

lib/utils/check-root-user.js
```diff
@@ -2,20 +2,53 @@
...
 const chalk = require('chalk');
+const fs = require('fs');
+const includes = require('lodash/includes');
 
-function checkRootUser() {
-    // Skip if we're on windows
-    if (os.platform() !== 'linux') {
+const isRootInstall = function isRootInstall() {
+    const path = require('path');
+
```
acburdine: Could simplify this logic here and simply say:

```js
const cliFile = path.join(process.cwd(), '.ghost-cli');
return fs.existsSync(cliFile) && fs.statSync(cliFile).uid === 0;
```

PR #637, opened 2018-02-07
Improve error messages in extensions

AileenCGN:
refs #587 

Trying to get a better error handling for our extensions. Catches an error whenever there was no catch, but one needed, creates proper errors when they were just thrown.

### DIFF
--- a/extensions/mysql/index.js
+++ b/extensions/mysql/index.js
@@ -119,15 +119,14 @@ class MySQLExtension extends cli.Extension {
                     `IDENTIFIED WITH mysql_native_password AS '${result[0].password}';`
                 ).catch((error) => {
                     // User already exists, run this method again
-                    if (error.errno === 1396) {
+                    if (error.err && error.err.errno === 1396) {
                         this.ui.logVerbose('MySQL: user exists, re-trying user creation with new username', 'yellow');
                         return tryCreateUser();
                     }
 
-                    return Promise.reject(new cli.errors.CliError({
-                        message: `Creating new MySQL user errored with message: ${error.message}`,
-                        err: error
-                    }));
+                    error.message = `Creating new MySQL user errored with message: ${error.message}`;
+
+                    return Promise.reject(error);
                 });
             };
 
@@ -143,14 +142,10 @@ class MySQLExtension extends cli.Extension {
         }).catch((error) => {
             this.ui.logVerbose('MySQL: Unable to create custom Ghost user', 'red');
             this.connection.end(); // Ensure we end the connection
-            if (error instanceof cli.errors.CliError) {
-                return Promise.reject(error);
-            }
 
-            return Promise.reject(new cli.errors.CliError({
-                message: `MySQL user creation errored with message: ${error.message}`,
-                err: error
-            }));
+            error.message = `Creating new MySQL user errored with message: ${error.message}`;
+
+            return Promise.reject(error);
         });
     }
 
@@ -163,16 +158,27 @@ class MySQLExtension extends cli.Extension {
         }).catch((error) => {
             this.ui.logVerbose('MySQL: Unable either to grant permissions or flush privileges', 'red');
             this.connection.end();
-            return Promise.reject(new cli.errors.CliError({
-                message: `Granting database permissions errored with message: ${error.message}`,
-                err: error
-            }));
+
+            error.message = `Granting database permissions errored with message: ${error.message}`;
+
+            return Promise.reject(error);
         });
     }
 
     _query(queryString) {
         this.ui.logVerbose(`MySQL: running query > ${queryString}`, 'gray');
-        return Promise.fromCallback(cb => this.connection.query(queryString, cb));
+        return Promise.fromCallback(cb => this.connection.query(queryString, cb))
+            .catch((error) => {
+                if (error instanceof (cli.errors.CliError)) {
+                    return Promise.reject(error);
+                }
+
+                return Promise.reject(new cli.errors.CliError({
+                    message: error.message,
+                    context: queryString,
+                    err: error
+                }));
+            });
     }
 }
 
--- a/extensions/mysql/index.js
+++ b/extensions/mysql/index.js
@@ -76,7 +76,7 @@ class MySQLExtension extends cli.Extension {
             }
 
             return Promise.reject(new cli.errors.CliError({
-                message: 'Error trying to connenct to the MySQL database.',
+                message: 'Error trying to connect to the MySQL database.',
                 help: 'You can run `ghost config` to re-enter the correct credentials. Alternatively you can run `ghost setup` again.',
                 err: error
             }));
--- a/extensions/mysql/index.js
+++ b/extensions/mysql/index.js
@@ -75,7 +75,11 @@ class MySQLExtension extends cli.Extension {
                 }));
             }
 
-            return Promise.reject(error);
+            return Promise.reject(new cli.errors.CliError({
+                message: 'Error trying to connenct to the MySQL database.',
+                help: 'You can run `ghost config` to re-enter the correct credentials. Alternatively you can run `ghost setup` again.',
+                err: error
+            }));
         });
     }
 
@@ -120,7 +124,10 @@ class MySQLExtension extends cli.Extension {
                         return tryCreateUser();
                     }
 
-                    return Promise.reject(error);
+                    return Promise.reject(new cli.errors.CliError({
+                        message: `Creating new MySQL user errored with message: ${error.message}`,
+                        err: error
+                    }));
                 });
             };
 
@@ -136,7 +143,14 @@ class MySQLExtension extends cli.Extension {
         }).catch((error) => {
             this.ui.logVerbose('MySQL: Unable to create custom Ghost user', 'red');
             this.connection.end(); // Ensure we end the connection
-            return Promise.reject(new cli.errors.SystemError(`Creating new mysql user errored with message: ${error.message}`));
+            if (error instanceof cli.errors.CliError) {
+                return Promise.reject(error);
+            }
+
+            return Promise.reject(new cli.errors.CliError({
+                message: `MySQL user creation errored with message: ${error.message}`,
+                err: error
+            }));
         });
     }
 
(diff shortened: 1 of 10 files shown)


AileenCGN:
@kirrg001 || @acburdine this is ready for review!

extensions/systemd/get-uid.js
```diff
@@ -23,7 +23,9 @@ module.exports = function getUid(dir) {
         return uid;
     } catch (e) {
         if (!e.message.match(/no such user/i)) {
-            throw e;
+            const errors = require('../../lib').errors;
+            // throw the error here so it'll be logged
+            throw new errors.ProcessError(e);
```
AileenCGN: I thought it might be a better idea to use a `ProcessError` here, as it gets logged at least.

extensions/systemd/index.js
```diff
@@ -17,7 +17,15 @@ class SystemdExtension extends cli.Extension {
     }
 
     _setup(argv, ctx, task) {
-        const uid = getUid(ctx.instance.dir);
+        let uid;
+
+        // getUid returns either the uid or null, but can also throw an error
```
AileenCGN: `getUid` was always able to throw an error, which we didn't handle anywhere. Therefore this refactor (same [here](https://github.com/TryGhost/Ghost-CLI/pull/637/files#diff-1edc5bfa2e72eb7d9a548201e7eb9674R104)).

extensions/systemd/systemd.js
```diff
@@ -91,15 +91,25 @@ class SystemdProcessManager extends cli.ProcessManager {
             // Systemd prints out "inactive" if service isn't running
             // or "activating" if service hasn't completely started yet
             if (!e.message.match(/inactive|activating/)) {
-                throw e;
+                throw new cli.errors.ProcessError(e);
```
AileenCGN: Same here as above, thought it might be better to not just throw the original error, but log it properly in the log files.

extensions/mysql/index.js
```diff
@@ -120,7 +124,10 @@ class MySQLExtension extends cli.Extension {
                         return tryCreateUser();
                     }
 
-                    return Promise.reject(error);
+                    return Promise.reject(new cli.errors.CliError({
```
kirrg001: Hm this looks redundant, because line 151 puts out a very similar error message. Or is there any difference?

AileenCGN: This particular catch is taking care of the `CREATE USER` query, whereas the catch in line 151 will take care of everything else that goes wrong. The error produced here will just be passed. But I can remove it, when it is redundant.

extensions/systemd/index.js
```diff
@@ -41,7 +49,9 @@ class SystemdExtension extends cli.Extension {
             ghost_exec_path: process.argv.slice(0,2).join(' ')
         }), 'systemd service', serviceFilename, '/lib/systemd/system').then(
             () => this.ui.sudo('systemctl daemon-reload')
-        );
+        ).catch((error) => {
+            return Promise.reject(new cli.errors.ProcessError(error));
```
kirrg001: In my opinion both `ctx.instance.template` and `this.ui.sudo` should return a custom error already.
So nobody has to handle errors if calling both units.

acburdine: Well, in some cases you might want to catch the error manually from sudo and do something with it. For instance, we catch any error from the sudo letsencrypt generation command via acme so that we can check the stderr. That would still be possible if `ui.sudo` caught and re-threw the error as a ProcessError, but it means additional logic.

At any rate, probably a subject for a later PR 😄

extensions/mysql/index.js
acburdine: I wonder if it's worth making the `_query` method handle the error instead and attach something to the error object based on the query string?

extensions/nginx/index.js
```diff
@@ -198,8 +208,12 @@ class NginxExtension extends cli.Extension {
                     confFile,
                     '/etc/nginx/sites-available'
                 ).then(
-                    () => this.ui.sudo(`ln -sf /etc/nginx/sites-available/${confFile} /etc/nginx/sites-enabled/${confFile}`)
-                );
+                    () => {
```
acburdine: quick formatting comment - this can still be a simple arrow function: `() => this.ui.sudo(...)`. If we for sure want to make it a normal arrow function (e.g. `() => { return this.ui.sudo(...); }`) can we move it up like this:

```js
.then(() => {
  return this.ui.sudo();
}).catch()

File: extensions/nginx/index.js [hunk 7]
```diff
@@ -222,7 +236,10 @@ class NginxExtension extends cli.Extension {
                     this.ui.sudo(`rm -f /etc/nginx/sites-available/${confFile}`),
                     this.ui.sudo(`rm -f /etc/nginx/sites-enabled/${confFile}`)
                 ]).catch(
-                    () => Promise.reject(new cli.errors.SystemError('Nginx config file link could not be removed, you will need to do this manually.'))
+                    (error) => Promise.reject(new cli.errors.CliError({
+                        message: `Nginx config file link could not be removed, you will need to do this manually for ${confFile}.`,
```
acburdine: this should maybe be a process error instead? Also, `confFile` will be the simple configuration filename, might be worth directing the user to remove the file in `/etc/sites-available` or `/etc/sites-enabled`.

extensions/nginx/index.js
```diff
@@ -234,7 +251,10 @@ class NginxExtension extends cli.Extension {
                     this.ui.sudo(`rm -f /etc/nginx/sites-available/${sslConfFile}`),
                     this.ui.sudo(`rm -f /etc/nginx/sites-enabled/${sslConfFile}`)
                 ]).catch(
-                    () => Promise.reject(new cli.errors.SystemError('SSL config file link could not be removed, you will need to do this manually.'))
+                    (error) => Promise.reject(new cli.errors.CliError({
```
acburdine: Same comment as above

extensions/systemd/index.js
```diff
@@ -17,7 +17,15 @@ class SystemdExtension extends cli.Extension {
     }
 
     _setup(argv, ctx, task) {
-        const uid = getUid(ctx.instance.dir);
+        let uid;
+
+        // getUid returns either the uid or null, but can also throw an error
+        try {
+            uid = getUid(ctx.instance.dir);
+        } catch (e) {
```
acburdine: Since we're catching the error and ignoring it, the change to getUID above (throwing a ProcessError instead) won't have any effect.

Not sure that anything really _needs_ to be changed here, but thought it worth mentioning.

extensions/nginx/index.js
AileenCGN: 🙈 Oops... Forgot to make it a normal arrow function again. I only changed it to log some stuff for myself...

extensions/nginx/index.js
AileenCGN: Yeah, I know... I also think it should be a `ProcessError`, but our usage of `ProcessError` doesn't include custom help text. That's why I picked the default `CliError`. 

Certainly makes sense to lead the user to the location of `confFile` for sure 👍

extensions/systemd/index.js
AileenCGN: I tried it without the `try` and `catch` after the changes and it was breaking for me... Can you think of a better way to solve this?

extensions/nginx/index.js
AileenCGN: Maybe it makes sense to submit another PR that improves the `ProcessError` to deal with custom properties?

extensions/mysql/index.js
AileenCGN: I'm not sure how I could do this with the `Promise.fromCallback` method? Can you give me a hint? Because I generally think that your suggestion is a very good idea.

acburdine: `Promise.fromCallback` returns a promise, so in the _query method you can just do:
```js
Promise.fromCallback(cb => ...).catch((error) => {
  // do stuff
});
```

extensions/systemd/get-uid.js
acburdine: Now that I'm thinking about it - is it worth checking the message here at all? could just return null regardless of the error message.

extensions/mysql/index.js
AileenCGN: This is done and pushed... Hope I did it right...

extensions/systemd/get-uid.js
AileenCGN: > Now that I'm thinking about it - is it worth checking the message here at all? could just return null regardless of the error message. (e.g. as was done in #659)

I agree and also changed the util in #659 to not throw any error. I explained it [here](https://github.com/TryGhost/Ghost-CLI/pull/659#discussion_r174345686) and also in the comment in the codebase [here](https://github.com/TryGhost/Ghost-CLI/pull/659/files#diff-e8c19ddf30935059861a1177570bd9e2R18).

I will refactor the code in the systemd util to do the same.

extensions/nginx/index.js
AileenCGN: I added a task to the existing improve error handling issue #587

AileenCGN:
@acburdine Can you give this another round? I updated the code as requested.

extensions/nginx/index.js
```diff
@@ -115,7 +125,10 @@ class NginxExtension extends cli.Extension {
                 return Promise.fromNode(cb => dns.lookup(parsedUrl.hostname, {family: 4}, cb)).catch((error) => {
                     if (error.code !== 'ENOTFOUND') {
                         // Some other error
-                        return Promise.reject(error);
+                        return Promise.reject(new cli.errors.CliError({
+                            message: `Error trying to looking DNS for '${parsedUrl.hostname}'`,
```
acburdine: slight wording fix - should be `Error trying to lookup DNS for '${parsedUrl.hostname}'`

AileenCGN: Oopsi...

extensions/mysql/index.js
```diff
@@ -75,7 +75,11 @@ class MySQLExtension extends cli.Extension {
                 }));
             }
 
-            return Promise.reject(error);
+            return Promise.reject(new cli.errors.CliError({
+                message: 'Error trying to connenct to the MySQL database.',
```
acburdine: /s/connenct/connect 😉

PR #647, opened 2018-02-16

vikaspotluri123:
Everything that hasn't been updated wasn't updated either because of a major version number or dropped v4 support


acburdine:
@vikaspotluri123 👍 for this, except find-plugins will need to be pinned to 1.1.5. Version 1.1.6 dropped node 4 support inadvertently.

PR #650, opened 2018-02-18
feat: make ghost (stop | start |  restart) more graceful

### LINKED ISSUE #649: Ghost start / stop / restart should fail gracefully
At the moment, running these commands:

`ghost stop` when an instance is stopped
`ghost start` when an instance is started
`ghost restart` when an instance is stopped

results in an ugly error. Since the purpose of the command in every case is already fulfilled, ghost should gracefully fail (i.e. `Ghost is already running! Nothing to do here`) and exit

vikaspotluri123:
closes #649 

- `ghost stop` notifies user and exits if instance is already stopped
- `ghost start` notifies user and informs of `ghost ls` command for more information
- `ghost restart` notifies user and runs start command if instance is not running already

There's another PR coming up to finish the minor refactor of the restart command!

### DIFF
--- a/lib/commands/restart.js
+++ b/lib/commands/restart.js
@@ -6,11 +6,12 @@ class RestartCommand extends Command {
         const instance = this.system.getInstance();
 
         if (!instance.running()) {
-            return Promise.reject(new Error('Ghost instance is not currently running.'));
+            const StartCommand = require('./start');
+            this.ui.log('Ghost instance is not running! Starting...', 'yellow');
+            return this.runCommand(StartCommand);
         }
 
         instance.loadRunningEnvironment(true);
-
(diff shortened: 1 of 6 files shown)


lib/commands/restart.js
```diff
@@ -6,11 +6,12 @@ class RestartCommand extends Command {
         const instance = this.system.getInstance();
 
         if (!instance.running()) {
-            return Promise.reject(new Error('Ghost instance is not currently running.'));
+            const StartCommand = require('./start');
+            this.ui.log('Ghost instance is not running! Starting...', 'yellow');
+            return this.runCommand(StartCommand, {quiet: true});
```
acburdine: the quiet option should be removed here imo - we want the start command to output the "Ghost instance has been started" message after Ghost starts.

vikaspotluri123:
Done 👍

PR #651, opened 2018-02-18
Refactor restart tests

vikaspotluri123:
(kind of) depends on #650 

- Remove unneeded class-based stubs
  - Was originally introduced in 650 to reduce unnecessary requires


acburdine:
looks like this needs a rebase 😉

vikaspotluri123:
Done!

PR #655, opened 2018-02-19
[WIP] chore(links): normalize styles for all links

vikaspotluri123:
Nothing's super set in stone; there are multiple styles applied to links, and this PR normalizes it to blue underlined (in that order)

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -141,10 +141,14 @@ const bootstrap = {
             process.exit(1);
         }
 
+        // Super lazy require!
+        const chalk = require('chalk');
+        const prettyDocsURL = chalk.blue.underline('https://docs.ghost.org/v1/docs/ghost-cli');
+
         // Yargs magic
         // TODO: evaluate if `wrap` is actually needed?
         yargs.wrap(Math.min(100, yargs.terminalWidth()))
-            .epilogue('For more information, see our docs at https://docs.ghost.org/v1/docs/ghost-cli')
+            .epilogue(`For more information, see our docs at ${prettyDocsURL}`)
             .option('D', {
                 alias: 'development',
                 describe: 'Run in development mode',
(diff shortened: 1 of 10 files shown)


AileenCGN:
Hey @vikaspotluri123 

We found that this blue is hardly readable sometimes and therefore went for green as the desired link colour, hence why it's green in the newer files (still needs to be changed in many places (e. g. [here](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/ui/index.js#L340)) Maybe it's just the setup in my shell, but it was barely readable for me.

I think in general, we should be more consistent for this, so this PR is definitely is the right approach already.

@kirrg001 @acburdine Can we do a general specification on how our logs should be rendered?

- Type of error -> red
- Error message: "Message:"/"Stack:" -> yellow, text -> white
- help -> green
- debug information -> yellow

Did I miss something? Any suggestions?

vikaspotluri123:
Yeah, I wasn't super sure what color to use so I just used the one I remember being most prominent, but I was definitely expecting some changes to be required 😄 

The default shells I use also make blue quite illegible on black, so no complaints with changing it here!

We might change help to light blue or cyan since green implies success 🤔

vikaspotluri123:
Maybe do `inverse` for code blocks 🤔

AileenCGN:
> We might change help to light blue or cyan since green implies success

Yeah, agree. I will check how we're doing the colours within Ghost core. Makes probably sense to keep it the same.

vikaspotluri123:
I'm guessing we should also get feedback from @peterzimon 😅

AileenCGN:
> I'm guessing we should also get feedback from @peterzimon

Yes, that is definitely something we can tackle in of our clean up weeks.

vikaspotluri123:
Sounds good - when we get to it should we add a wrapper util for the different types?

PR #657, opened 2018-02-22
feat(doctor): prevent installations with ghost user

AileenCGN:
refs #47

- Adds a check to the `install` category
- Check is disabled for local installs and platforms that are not linux
- Check will prevent user from installation, if the `ghost` user exists and user is logged in as such
- Adds tests. Still 💯

Outcome:
image

### DIFF
--- a/lib/commands/doctor/checks/logged-in-user.js
+++ b/lib/commands/doctor/checks/logged-in-user.js
@@ -1,0 +1,28 @@
+'use strict';
+const errors = require('../../../errors');
+const chalk = require('chalk');
+const ghostUser = require('../../../utils/use-ghost-user');
+
+const taskTitle = 'Checking logged in user';
+
+function loggedInUser() {
+    const uid = process.getuid();
+    const ghostStats = ghostUser.getGhostUid();
+
+    if (ghostStats && ghostStats.uid === uid) {
+        throw new errors.SystemError({
+            message: 'You can\'t install Ghost with a user called `ghost`. Please use a different user(name).',
+            help: `${chalk.green('https://docs.ghost.org/docs/install#section-create-a-new-user')}`,
+            task: taskTitle
+        });
+    }
+
+    return;
+}
+
+module.exports = {
+    title: taskTitle,
+    task: loggedInUser,
+    enabled: (ctx) => !ctx.local && !(ctx.instance && ctx.instance.process.name === 'local') && ctx.system.platform.linux && !(ctx.argv && ctx.argv.process === 'local'),
+    category: ['install']
+};
--- a/lib/commands/doctor/checks/index.js
+++ b/lib/commands/doctor/checks/index.js
@@ -1,5 +1,6 @@
 'use strict';
 const nodeVersion = require('./node-version');
+const loggedInUser = require('./logged-in-user');
 const loggedInGhostUser = require('./logged-in-ghost-user');
 const loggedInUserOwner = require('./logged-in-user-owner');
 const installFolderPermissions = require('./install-folder-permissions');
@@ -13,6 +14,7 @@ const checkMemory = require('./check-memory');
 
 module.exports = [
     nodeVersion,
+    loggedInUser,
     loggedInGhostUser,
     loggedInUserOwner,
     installFolderPermissions,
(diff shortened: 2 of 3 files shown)


lib/commands/doctor/checks/logged-in-user.js
```diff
@@ -0,0 +1,39 @@
...
+
+function loggedInUser() {
+    const uid = process.getuid();
+    let ghostuid;
+
+    try {
+        ghostuid = execa.shellSync('id -u ghost').stdout;
+    } catch (e) {
+        // CASE: the ghost user doesn't exist yet, all good in the hood
+        return Promise.resolve();
```
acburdine: this can just be `return` -> there's no async behavior anywhere in the function so no need to make it async.

lib/commands/doctor/checks/logged-in-user.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    ghostuid = parseInt(ghostuid);
+
+    if (ghostuid === uid) {
+        return Promise.reject(new errors.SystemError({
```
acburdine: same here, this can just be `throw`

lib/commands/doctor/checks/logged-in-user.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    return Promise.resolve();
+}
+
+module.exports = {
+    title: taskTitle,
+    task: loggedInUser,
+    enabled: (ctx) => !ctx.local && !(ctx.instance && ctx.instance.process.name === 'local'),
+    skip: () => os.platform() !== 'linux',
```
acburdine: should probably get the platform from `ctx.system.platform`, although I'm not 100% either on making it a "skip if not linux". imo it might make more sense to flip the enabled & skip functions?

AileenCGN: Hmm... I changed it to `ctx.system.platform` (should have known that, as I implemented it 🙈), but I'm not sure about flipping it... In `system-stack` which is also an install task, we have the enabled set the same. See [here](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/doctor/checks/system-stack.js#L66).

lib/commands/doctor/checks/logged-in-user.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    return;
+}
+
+module.exports = {
+    title: taskTitle,
+    task: loggedInUser,
+    enabled: (ctx) => !ctx.local && !(ctx.instance && ctx.instance.process.name === 'local'),
+    skip: (ctx) => !ctx.system.platform.linux,
```
acburdine: this condition should be moved into `enabled` I think.

lib/commands/doctor/checks/logged-in-user.js
```diff
@@ -0,0 +1,37 @@
...
+const errors = require('../../../errors');
+const chalk = require('chalk');
+
+const taskTitle = 'Checking logged in user';
+
+function loggedInUser() {
+    const uid = process.getuid();
+    let ghostuid;
+
+    try {
```
AileenCGN: I will change this to use the new `getGhostUid` fn [here](https://github.com/TryGhost/Ghost-CLI/pull/659/files#diff-e8c19ddf30935059861a1177570bd9e2R6) as soon as #659 is merged.

lib/commands/doctor/checks/logged-in-user.js
```diff
@@ -0,0 +1,37 @@
...
+        });
+    }
+
+    return;
+}
+
+module.exports = {
+    title: taskTitle,
+    task: loggedInUser,
+    enabled: (ctx) => !ctx.local && !(ctx.instance && ctx.instance.process.name === 'local') && ctx.system.platform.linux,
```
acburdine: might be worth also checking if `argv.process === 'local'` -> that way if you do `ghost install --process local`, this check will be skipped.

AileenCGN: Good call!!! Changes are pushed! Ready for 🚀

PR #659, opened 2018-02-23
feat(doctor): check current user and dir owner

AileenCGN:
refs #47

- Adds a check to the `start` and `update` category
- Check is disabled for local installs and platforms that are not linux
- Check will prevent from proceeding if:
	- if a user named `ghost` exists and if it's currently used and also the owner of the content folder
	- if the uid of the current user is not the same uid of the ghost installation folder, and the owner of the installation folder is also not in the same group as the current user
- Adds tests. Still 💯

image
^ logged in as ghost and ghost is also used as system user


lib/commands/doctor/checks/logged-in-user-owner.js
```diff
@@ -0,0 +1,76 @@
...
+    const dirStats = fs.lstatSync(path.join(process.cwd()));
+    const contentDirStats = fs.lstatSync(path.join(process.cwd(), 'content'));
+
+    // CASE 1: check if ghost user exists and if it's currently used
+    if (ghostStats && ghostStats.uid && ghostStats.uid === uid) {
+        // The ghost user might have been set up on the system and also used,
+        // but only when it owns the content folder, it's an indication that it's also used
+        // as the linux user and shall not be used as current user.
+        if (contentDirStats.uid === ghostStats.uid) {
+            return Promise.reject(new errors.SystemError({
```
acburdine: similar comments to #657 - this doesn't need to be a promise.

lib/commands/doctor/checks/logged-in-user-owner.js
```diff
@@ -0,0 +1,76 @@
...
+        if (gid.indexOf(dirStats.gid) < 0) {
+            return Promise.reject(new errors.SystemError({
+                message: `Your current user is not the owner of the Ghost directory and also not part of the same group.
+Please log in with the user that owns the directory or add your user to the same group.`,
+                help: `${chalk.green('https://docs.ghost.org/docs/install#section-create-a-new-user')}`,
+                task: taskTitle
+            }));
+        }
+        // Yup current user is not the owner, but in the same group, so just show a warning
+        ctx.ui.log('The current user is not the owner of the Ghost directory. This might cause problems.');
```
acburdine: does this need to be different color text since it's a warning?

lib/commands/doctor/checks/logged-in-user-owner.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    return Promise.resolve();
+}
+
+module.exports = {
+    title: taskTitle,
+    task: loggedInUserOwner,
+    enabled: (ctx) => ctx.instance && ctx.instance.process.name !== 'local',
+    skip: () => os.platform() !== 'linux',
```
acburdine: should probably flip the skip and enabled functions here, similar to #657

lib/commands/doctor/checks/logged-in-user-owner.js
```diff
@@ -0,0 +1,76 @@
...
+const os = require('os');
+const execa = require('execa');
+const path = require('path');
+const errors = require('../../../errors');
+const chalk = require('chalk');
+const fs = require('fs');
+
+const taskTitle = 'Checking logged in user and directory owner';
+
+function checkGhostUser() {
```
acburdine: is this worth throwing in a util? Seems that it's similar to https://github.com/TryGhost/Ghost-CLI/blob/master/lib/utils/use-ghost-user.js and there could be some common code 😄

AileenCGN: I could refactor the existing util to have two different fn maybe? One to get the ghost uid and gid and one to check if the ghost user should be used because he owns the content directory?

acburdine: Yeah that’s what I was thinking 🙂

lib/commands/doctor/checks/logged-in-user-owner.js
AileenCGN: In `system-stack` which is also an install task, we have the enabled set the same. See [here](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/doctor/checks/system-stack.js#L66). Is there a reason for that? Shall we do it the same everywhere?

lib/commands/doctor/checks/logged-in-user-owner.js
AileenCGN: This is done and updated!

lib/utils/use-ghost-user.js
```diff
@@ -16,21 +16,51 @@ module.exports = function useGhostUser(contentDir) {
         ghostuid = execa.shellSync('id -u ghost').stdout;
         ghostgid = execa.shellSync('id -g ghost').stdout;
     } catch (e) {
+        // CASE: the ghost user doesn't exist, hence can't be used
         if (!e.message.match(/no such user/i)) {
             throw new errors.ProcessError(e);
         }
 
-        return false;
+        return false
```
acburdine: need to fix the lint rules, but this needs a semicolon I think?

lib/utils/use-ghost-user.js
```diff
@@ -16,21 +16,51 @@ module.exports = function useGhostUser(contentDir) {
...
+
+const shouldUseGhostUser = function shouldUseGhostUser(contentDir) {
+    let ghostUser;
+
+    if (os.platform() !== 'linux') {
+        return false;
+    }
+
+    // get the ghost uid and gid
+    try {
```
acburdine: since this is just catching and re-throwing the error, we can just remove the try/catch block altogether.

lib/commands/doctor/checks/logged-in-user-owner.js
```diff
@@ -0,0 +1,61 @@
...
+const taskTitle = 'Checking logged in user and directory owner';
+
+function loggedInUserOwner(ctx) {
+    const uid = process.getuid();
+    const gid = process.getgroups();
+    let ghostStats;
+    const dirStats = fs.lstatSync(path.join(process.cwd()));
+    const contentDirStats = fs.lstatSync(path.join(process.cwd(), 'content'));
+
+    try {
```
acburdine: unnecessary try/catch block.

lib/commands/doctor/checks/logged-in-user-owner.js
acburdine: Well, in the case of the system-stack check the priorities are a bit different, esp. because the system stack check runs on install whereas this is a start/update task 😄

If anything the skip function should be made the `enable` function, and the current enabled function should be removed. Even if you're using the local process manager on linux, you could still run into issues with file permissions so it's probably still worth checking.

lib/commands/doctor/checks/logged-in-user-owner.js
```diff
@@ -0,0 +1,61 @@
...
+        ghostStats = ghostUser.getGhostUid();
+    } catch (e) {
+        throw e;
+    }
+
+    // CASE 1: check if ghost user exists and if it's currently used
+    if (ghostStats && ghostStats.uid && ghostStats.uid === uid) {
```
acburdine: Not 100% on adding this check....seems like this would be better moved to a systemd extension check (or at least a separate check)

lib/commands/doctor/checks/logged-in-user-owner.js
```diff
@@ -0,0 +1,61 @@
...
+Please log in with the user that owns the directory or add your user to the same group.`,
+                help: `${chalk.green('https://docs.ghost.org/docs/install#section-create-a-new-user')}`,
+                task: taskTitle
+            });
+        }
+        // Yup current user is not the owner, but in the same group, so just show a warning
+        ctx.ui.log(`${chalk.yellow('The current user is not the owner of the Ghost directory. This might cause problems.')}`);
+    }
+
+    return;
```
acburdine: unnecessary return

lib/commands/doctor/checks/logged-in-user-owner.js
AileenCGN: > Not 100% on adding this check

Can you explain this to me a bit further? This check is trying to prevent that a user is running commands as a logged in `ghost` user.

lib/utils/use-ghost-user.js
```diff
@@ -16,6 +16,7 @@ module.exports = function useGhostUser(contentDir) {
         ghostuid = execa.shellSync('id -u ghost').stdout;
         ghostgid = execa.shellSync('id -g ghost').stdout;
     } catch (e) {
+        // CASE: the ghost user doesn't exist, hence can't be used
         if (!e.message.match(/no such user/i)) {
             throw new errors.ProcessError(e);
```
AileenCGN: @acburdine I am actually not 100% sure if we should throw an error here at all. I can only see two reasons when the command wouldn't work: 
1. Command is not known by Linux
2. User doesn't exist

The first case is not very likely to happen, I assume. And the second case would exactly be the result we want. `false` if the user doesn't exist. 

I will change it for now...

acburdine: That's a fair point.

lib/commands/doctor/checks/logged-in-user-owner.js
acburdine: So the idea behind doctor checks is to have one thing per check, right? Checking whether or not the current user is `ghost` is something entirely different from checking whether or not the current directory is owned by the current running user. It complicates the `enabled`/`skip` checks as well, because for the local process manager, it's good to check whether or not the current user owns the current folder, but for the local process manager the user _can_ be ghost.

AileenCGN: Alright. Gotcha! 👍 I think I had a little false thinking in my brain, assuming that those checks belong together. I split them now in two separate checks and also skipped the `ghost-user` check when the local process manager is used.

AileenCGN:
@acburdine I think this is also ready for another review round!

AileenCGN:
@acburdine ping

AileenCGN:
Please do!!!

Issue #660, opened 2018-02-24
Ghost CLI Validating config undefined.


techayu:
Welcome to Ghost-CLI's GitHub repo! 👋🎉

Do you need help or have a question? Please come chat in Slack: https://slack.ghost.org/ 👫.

Docs: https://docs.ghost.org 📖.

Please be aware that the team behind the Ghost CLI only supports the recommended stack: https://github.com/TryGhost/Ghost-cli#recommended-stack.

## This issue is a

* [x] Bug Report
* [x] Feature Request

Please fill out this template, deleting where appropriate:

### Summary

Bugs reports: search GitHub for existing issues & check the docs: https://docs.ghost.org/docs/troubleshooting. If you're still stuck, please provide a quick summary of the problem, steps to reproduce, and full tech details including logs.

Feature requests: describe the feature and make a clear, detailed case for adding it.

### Steps to Reproduce (for a bug report)

I have installed ghost many times on my vps successfully without any error on same operating system. This time also i am using ubuntu 16.04 64 bit.

Any other info e.g. Why do you consider this to be a bug? What did you expect to happen instead?

### Log file (for a bug report)

All errors end with "Additional log info available in: [filepath]".

Use the command `cat [filepath]` to read the log, then copy & paste the contents here:

```
Paste log file here
aDebug Information:
    OS: Ubuntu, v16.04
    Node Version: v6.13.0
    Ghost-CLI Version: 1.5.2
    Environment: production
    Command: 'ghost install'
An error occurred.
Message: 'Command failed: /bin/sh -c systemctl is-active ghost_localhost
Failed to connect to bus: No such file or directory

'

Stack: Error: Command failed: /bin/sh -c systemctl is-active ghost_localhost
Failed to connect to bus: No such file or directory


    at makeError (/usr/lib/node_modules/ghost-cli/node_modules/execa/index.js:169:9)
    at module.exports.sync (/usr/lib/node_modules/ghost-cli/node_modules/execa/index.js:338:15)
    at handleShell (/usr/lib/node_modules/ghost-cli/node_modules/execa/index.js:116:9)
    at Function.module.exports.shellSync (/usr/lib/node_modules/ghost-cli/node_modules/execa/index.js:361:43)
    at SystemdProcessManager.isRunning (/usr/lib/node_modules/ghost-cli/extensions/systemd/systemd.js:88:19)
    at Instance.running (/usr/lib/node_modules/ghost-cli/lib/instance.js:120:34)
[...]

PR #668, opened 2018-03-11
Fix(Systemd): fix execa command (issue #660)

uadp:
Regarding #660 Issue

- Add .service file extension
- Add permission for execa command (sudo)
- Fix mismatch error catch execa

### DIFF
--- a/extensions/systemd/systemd.js
+++ b/extensions/systemd/systemd.js
@@ -8,7 +8,7 @@ const chalk = require('chalk');
 
 class SystemdProcessManager extends cli.ProcessManager {
     get systemdName() {
-        return `ghost_${this.instance.name}`;
+        return `ghost_${this.instance.name}.service`;
     }
 
     get logSuggestion() {
@@ -60,12 +60,12 @@ class SystemdProcessManager extends cli.ProcessManager {
 
     isEnabled() {
         try {
-            execa.shellSync(`systemctl is-enabled ${this.systemdName}`);
+            execa.shellSync(`sudo systemctl is-enabled ${this.systemdName}`);
             return true;
-        } catch (e) {
+        } catch (err) {
             // Systemd prints out "disabled" if service isn't enabled
             // or "failed to get unit file state" if something else goes wrong
-            if (!e.message.match(/disabled|Failed to get unit file state/)) {
+            if (!err.message.match(/disabled|Failed to get unit file state/)) {
                 throw e;
             }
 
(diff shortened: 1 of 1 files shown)


extensions/systemd/systemd.js
```diff
@@ -60,12 +60,12 @@ class SystemdProcessManager extends cli.ProcessManager {
 
     isEnabled() {
         try {
-            execa.shellSync(`systemctl is-enabled ${this.systemdName}`);
+            execa.shellSync(`sudo systemctl is-enabled ${this.systemdName}`);
```
vikaspotluri123: Sudo commands should be run using `this.ui.sudo({{command_to_run}})`

extensions/systemd/systemd.js
```diff
@@ -85,12 +85,12 @@ class SystemdProcessManager extends cli.ProcessManager {
 
     isRunning() {
         try {
-            execa.shellSync(`systemctl is-active ${this.systemdName}`);
+            execa.shellSync(`sudo systemctl is-active ${this.systemdName}`);
```
vikaspotluri123: Use sudo helper

uadp:
Well i'm not too well-versed with nodejs. i'm still learning anyway.

i have tried with sudo helper, but the result didn't same. (or i didn't capable to make it)
i haven't idea to use sudo helper with same return as shellAsync (Returns the same result object as child_process.spawnSync), i guess we must add some lines in process-manager.js.

But adding sudo and fixing execa error catch, it's easy way to resolve the issue :smiley: 
and it will help new user to install ghost smoothly with recommend stack.

extensions/systemd/systemd.js
acburdine: As much as it would be nice to do that here - the sudo helper returns a promise, and as such won't work without a decent amount of refactoring in other parts of the code 😕

vikaspotluri123: that's right, oops 🙊 

The very minimum is to add logging in that case, right? 🤔

uadp:
@acburdine Regarding this error output

`Message: 'Command failed: /bin/sh -c systemctl is-active ghost_localhost
Failed to connect to bus: No such file or directory`

it seem the command didn't have enough permission to run systemctl .service file.
since ghost command didn't run as root user.

It short cut to solve the issue,
solving with modification server side systemd is more complicated.

acburdine:
I'm gonna look into exactly how difficult it would be to allow `isRunning` to return a promise. I know in other process managers it could simplify some things.

AileenCGN:
I submitted a PR that will change those two commands to run as sudo: #672

acburdine:
Closed by #672.

PR #681, opened 2018-03-21
Ensure `content/settings` folder

AileenCGN:
refs TryGhost/Ghost#9528

- adds `content/settings` folder to install process (for new installations)
- adds a core migration that makes sure, `content/settings` exists (for older installations)

----

## Todos:
- [ ] ~~create default `routes.yaml` file on install and migration. Question: where to get the default values from?~~
- [ ] make sure it works fine with the Ghost changes
- [x] define the version for the migration

### DIFF
--- a/lib/migrations.js
+++ b/lib/migrations.js
@@ -1,3 +1,15 @@
 'use strict';
 
-module.exports = [];
+function ensureSettingsFolder() {
+    const cwd = process.cwd();
+    const fs = require('fs-extra');
+    const path = require('path');
+
+    fs.ensureDirSync(path.resolve(cwd, 'content', 'settings'));
+}
+
+module.exports = [{
+    before: '1.7.0',
+    title: 'Create content/settings directory',
+    task: ensureSettingsFolder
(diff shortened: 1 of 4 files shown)


kirrg001:
@acburdine 

It would be good to ship this asap. When is the next release planned?
It will ensure that the content folder get's a new sub folder.

And if we have released this, we can then release https://github.com/TryGhost/Ghost/pull/9534 with a new CLI engine bump in Ghost. 

Currently we throw

> Zip file contains a Ghost version incompatible with this version of the CLI.

Would be great to extend this error a bit and tell the user what todo cc @AileenCGN.

kirrg001:
@AileenCGN Why is this on HOLD?

> create default routes.yaml file on install and migration. Question: where to get the default values from?

No.

> define the version for the migration

@acburdine

AileenCGN:
@acburdine Version is now set to 1.7.0

acburdine:
Works - thanks @AileenCGN 😄

PR #684, opened 2018-03-25
fix(setup): fix start/no-start option

### LINKED ISSUE #615: `--no-start` flag gets ignored when running `ghost install local`
[placeholder] When running `ghost install local --no-start`, Ghost starts anyway.

acburdine:
closes #615
- respect no-start in local installs
- fix ui.confirm usage

### DIFF
--- a/lib/commands/setup.js
+++ b/lib/commands/setup.js
@@ -58,7 +58,7 @@ class SetupCommand extends Command {
                 argv.dbpath = path.join(process.cwd(), 'content/data/ghost-local.db');
             }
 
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
What were the changes w/ `ui.confirm`?

acburdine:
@vikaspotluri123 so instead of passing back an object with {yes: <whether or not thing was confirmed>) the ui.confirm option was updated to return just a straight boolean - made some things simpler. Missed this particular bit of code initially though when I went in and changed the method :)

PR #688, opened 2018-03-27
chore(deps): switch from os to systeminformation for free-memory availability

### LINKED ISSUE #687: Memory check reads wrong memory
With https://github.com/TryGhost/Ghost-CLI/commit/a852cddc6f2f5c3e7b57d9205f23f6b4329439e5 we introduced the memory check for memory heavy tasks.

This is using `os.freemem()`, which apparently is based on the wrong value. See the Slack conversation [here](https://ghost.slack.com/archives/C04CUS2R3/p1522139085000093).

There's a reference on how to possibly fix it [here](https://stackoverflow.com/questions/20578095/node-js-get-actual-memory-usage-as-a-percent).

vikaspotluri123:
refs #540 , closes #687 

systeminformation has pretty good coverage of the different OSes when it comes to checking memory availability


lib/commands/doctor/checks/check-memory.js
```diff
@@ -1,16 +1,19 @@
...
 const MIN_MEMORY = 150;
 
 function checkMemory() {
-    const availableMemory = os.freemem() / MB_IN_BYTES;
-    if (availableMemory < MIN_MEMORY) {
-        return Promise.reject(new SystemError(`Ghost recommends you have at least ${MIN_MEMORY} MB of memory available for smooth operation. It looks like you have ${parseInt(availableMemory)} MB available.`));
-    }
-    return Promise.resolve();
+    return sysinfo.mem().then((memoryInfo) => {
+        return (memoryInfo.free / MB_IN_BYTES);
```
acburdine: this should probably be `memoryInfo.available`? memoryInfo.free seems to be the same thing that os.freemem returns, whereas available returns a more holistic number that includes available buffer & cache memory.

Also, there's no need for the additional promise code here, you could just do:
```js
return sysinfo.mem().then((memoryInfo) => {
  const availableMemory = memoryInfo.available / MB_IN_BYTES;

  if (availableMemory < MIN_MEMORY) {
    return Promise.reject(); // ...
  }

  return Promise.resolve();
});

vikaspotluri123: Oops, I saw that free was set (in systeminformation source) and jumped the gun 😬

PR #693, opened 2018-04-04
feat(config): add 'get key' and 'set key value' support

vikaspotluri123:


acburdine:
I need to test this a bit more before merging - also there might be some fancy things we can do here with the yargs `positional` [function](https://github.com/yargs/yargs/blob/master/docs/api.md#positionalkey-opt).

Will definitely get this change in the next version though!

acburdine:
Gonna do a bit more investigating into this. It might be possible to accomplish the same changes via yargs subcommands, which might make things a bit cleaner.

vikaspotluri123:
I did my best to implement it, but the issue I see (which you might have a better understanding of how to implement ) is when there are 2 arguments - it can either be `get key` or `key value` (with implicit set). So if yargs has the capability of conditional argument assignment, my argument reassignment could be removed

acburdine:
@vikaspotluri123 so it turns out my thought about subcommands was right - this approach seemed to work for me locally: https://github.com/acburdine/Ghost-CLI-1/commit/55fbf1187876f3c24f46cca7e7c7ea57c9508cd1

vikaspotluri123:
Yep, closing in favor of #731! Much more logical 😄

PR #695, opened 2018-04-04
Drop node v4 (EOL)

vikaspotluri123:
At the end of April, node v4 will reach end-of-life status, and v10 will gain Active LTS. Ghost's [supported node versions](https://docs.ghost.org/docs/supported-node-versions) are influenced by LTS, so this PR contains updates to the CLI's supported engines as well as remove legacy code v4 required.


vikaspotluri123:
Rebased 🎉

PR #703, opened 2018-04-11
fix(migrations): Use proper permissions to create content/settings folder

e-nomem:
When migrating from CLI version `1.6.0` to `1.7.0`, the new `content/settings` directory is not properly created since the `content` directory is owned by the `ghost` user and not my running user.

The fix here is to check the owner of the directory first and use `sudo` as appropriate.

It's getting late here so I'll come back for the failing tests later unless someone here wants to fix them up.

### DIFF
--- a/test/unit/migrations-spec.js
+++ b/test/unit/migrations-spec.js
@@ -2,26 +2,57 @@
 
 const expect = require('chai').expect;
 const sinon = require('sinon');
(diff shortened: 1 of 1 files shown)


AileenCGN:
@e-nomem Thank you very much!!! This is totally necessary! I will fix the tests and get this merged asap!

PR #728, opened 2018-05-24
Updated the folder check's help command to consider special permissions

snarlynarwhal:
Updated the folder check's help command to remove any special permissions (setuid, setgid and sticky bits) in addition to setting the proper standard permissions.


vikaspotluri123:
Tests are failing because you need to update the help command here:

`test/unit/commands/doctor/checks/folder-permissions-spec.js:44`
`test/unit/commands/doctor/checks/folder-permissions-spec.js:58`

Could you also add a comment which describes what the three zeros are for?

PR #731, opened 2018-05-25
feat(config): add config get & set subcommands

acburdine:
this is another approach to the feature added in #693 , using yargs subcommand support instead.

### DIFF
--- a/lib/commands/config/index.js
+++ b/lib/commands/config/index.js
@@ -5,6 +5,18 @@ const Command           = require('../../command');
 const advancedOptions   = require('./advanced');
 
 class ConfigCommand extends Command {
+    static configureSubcommands(commandName, commandArgs, extensions) {
+        return commandArgs.command({
+            command: 'get <key>',
+            describe: 'Get a specific value from the configuration file',
+            handler: (argv) => this._run(`${commandName} get`, argv, extensions)
+        }).command({
+            command: 'set <key> <value>',
+            describe: 'Set a specific value in the configuration file',
+            handler: (argv) => this._run(`${commandName} set`, argv, extensions)
+        });
+    }
+
     constructor(ui, system) {
         super(ui, system);
 
@@ -154,6 +166,7 @@ class ConfigCommand extends Command {
         } else if (key) {
             // setter
             this.instance.config.set(key, value).save();
+            this.ui.log(`Successfully set '${key}' to '${value}'`, 'green');
             return Promise.resolve();
         }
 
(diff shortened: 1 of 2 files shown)


lib/commands/config/index.js
```diff
@@ -5,6 +5,16 @@ const Command           = require('../../command');
 const advancedOptions   = require('./advanced');
 
 class ConfigCommand extends Command {
+    static configureSubcommands(commandName, commandArgs, extensions) {
+        return commandArgs.command({
```
vikaspotluri123: Do you know if it's possible to expose these options when running `ghost config help`?

lib/commands/config/index.js
```diff
@@ -5,6 +5,16 @@ const Command           = require('../../command');
...
+            command: 'get <key>',
+            handler: (argv) => this._run(`${commandName} get`, argv, extensions)
```
vikaspotluri123: What exactly is the reasoning behind renaming the command here? Is it just for convenience or does yargs  somehow use it?

vikaspotluri123:
I didn't think of chaining `.command` but now that I see it it makes a lot of sense!

lib/commands/config/index.js
acburdine: Hmm...I didn’t test it but it should have. I’ll look into it.

lib/commands/config/index.js
acburdine: So the reason that’s passed in is the debug output uses that command name to say the error occurred running command “x”. Probably not a _huge_ deal with these small commands, but it’s probably worth it to be consistent 🙂

vikaspotluri123: Got it - I guess it just feels weird to me that the command name is `ghost config (s|g)et`

lib/commands/config/index.js
acburdine: Looks like all I needed to do was add a description 😁

acburdine:
This should be good for a review now @AileenCGN  😄

kirrg001:
I execute `ghost config` in an empty folder, it tells me

> Working directory is not a recognisable Ghost installation.

Makes sense. Then I execute `ghost config get` in an empty folder, it shows me the whole help output 🤔 

Feels a little inconsistent? 

---

Furthermore, if i set a value, i am not getting any feedback like "Successfully updated the config value".

acburdine:
@kirrg001 the reason you're getting help output when you run `ghost config get` is because `ghost config get` requires a `key` to get a value for, and because you didn't pass one Yargs prints out help output.

acburdine:
ok @kirrg001 - the log has been added :)

Issue #726, opened 2018-05-23
Upgrade failure - unclear and unclean handling


ErisDS:
I've been upgrading a bunch of Ghost CLI managed blogs today. Most of the upgrades went fine, but one of them encountered a network error with npm. This will happen from time to time, and should be handled cleanly.

Instead, what happened next resulted in the blog being taken offline, which is totally not what I would have expected.

### Steps to Reproduce (for a bug report)

This is not really reproducible without faking a connection error, but here's what happened:

1. I ran `ghost update` everything was happy green ticks until...

```
✖ Downloading and updating Ghost to v1.23.0
A ProcessError occurred.

Message: Command failed: yarn install --no-emoji --no-progress
warning ghost@1.23.0: The engine "cli" appears to be invalid.
error An unexpected error occurred: "https://registry.yarnpkg.com/@tryghost/extract-zip/-/extract-zip-1.6.6.tgz: ESOCKETTIMEDOUT".

yarn install v1.6.0
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info There appears to be trouble with your network connection. Retrying...
info If you think this is a bug, please open a bug report with the information provided in "/var/www/ghost/versions/1.23.0/yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

Exit code: 1
```
(no additional output in logs)

2. I re-ran `ghost doctor`, just to be absolutely certain. All green.

3. I re-ran `ghost update`

```
✔ Checking for latest Ghost version
ℹ Downloading and updating Ghost [skipped]
Running sudo command: systemctl is-active ghost_mysite
Running sudo command: systemctl stop ghost_mysite
✔ Stopping Ghost
✔ Linking latest Ghost and recording versions
Running sudo command: /var/www/ghost/current/node_modules/.bin/knex-migrator-migrate --init --mgpath /var/www/ghost/current
✖ Running database migrations
A ProcessError occurred.

Message: The database migration in Ghost encountered an error.
[...]

PR #732, opened 2018-05-25
Ensure version folder is cleaned up in the event of a yarn install failure.

acburdine:
refs #726 
- if an error occurs during yarn install (called as part of ghost install or ghost update) we remove the version folder for the particular install so that future runs of `ghost update` don't treat the folder as successfully installed.


acburdine:
Looks like tests are failing because of this issue: https://github.com/yarnpkg/yarn/issues/5885 -> once it's resolved the tests should work again 😄

vikaspotluri123:
Looks like there were a couple of refs to the `env` fixture which weren't updated

PR #734, opened 2018-05-27
feat(flags): add global `dir` option

### LINKED ISSUE #538: Global directory command
The basic idea is to add the `--dir | -d` option as a global option, similar to `-D | --development` and `-V | --verbose`.

One of the goals for this (outside of convenience) is to abstract permission-checking out of specific commands so there's one less thing to worry about per-command, and hopefully will also help make debugging easier for people who are installing ghost on an established system.

My thoughts are:
- The default dir is `.`
- If the command is global, nothing is done
- Every command would have an additional `write` and `read` boolean option (my thoughts for defaults are `read: true, write: false`) and the permissions would be checked based on said options.

I wanted to let this sit here for a bit before I started implementation to see if anyone sees any issues with this, or feels something should be included.

vikaspotluri123:
Add ability to specify which directory Ghost should run in

refs + closes #538

WIP
- Should permissions even be checked? We already have improvements to Ghost doctor which included much more in-depth + variable permissions checks
- Tests 🙃

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -145,7 +145,10 @@ const bootstrap = {
         // TODO: evaluate if `wrap` is actually needed?
         yargs.wrap(Math.min(100, yargs.terminalWidth()))
             .epilogue('For more information, see our docs at https://docs.ghost.org/v1/docs/ghost-cli')
-            .option('D', {
+            .option('d', {
+                alias: 'dir',
+                describe: 'Folder to run command in'
+            }).option('D', {
                 alias: 'development',
                 describe: 'Run in development mode',
                 type: 'boolean'
--- a/lib/command.js
+++ b/lib/command.js
@@ -107,7 +107,42 @@ class Command {
      * @method run
      * @private
      */
-    static _run(commandName, argv, extensions) {
+    static _run(commandName, argv = {}, extensions) {
+        debug('running command prep');
+        // Set process title
+        process.title = `ghost ${commandName}`;
+
+        // Create CLI-wide UI & System instances
+        const ui = new UI({
+            verbose: argv.verbose,
+            allowPrompt: argv.prompt
+        });
+
+        // This needs to run before the installation check
+        if (argv.dir) {
+            debug('Directory specified, attempting to update');
+            const path = require('path');
+            const dir = path.resolve(argv.dir);
+            try {
+                if (this.ensureDir) {
+                    const {ensureDirSync} = require('fs-extra');
+                    ensureDirSync(dir);
+                }
+
+                process.chdir(dir);
+            } catch (error) {
+                /* istanbul ignore next */
+                const err = error.message || error.code || error;
+                ui.log(`Unable to use "${dir}" (error ${err}). Please fix the issue and try again.`, 'red', true);
+                process.exit(1);
+            }
+            /* istanbul ignore next */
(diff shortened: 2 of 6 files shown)


lib/command.js
```diff
@@ -108,7 +108,26 @@ class Command {
...
+            const dir = path.resolve(argv.dir);
+
+            if (this.ensureDir) {
+                fs.ensureDirSync(dir);
+            }
+
+            process.chdir(dir);
+        } catch (error) {
+            const chalk = require('chalk');
+            console.warn(chalk.yellow(`Error using "${argv.dir}" (${error.code || error.message || error}) - staying in"${process.cwd()}"`));
```
acburdine: couple things about this:

1) would it be possible to move this down further in the run? that way this could have access to the UI class instead of needing to do console.warn
2) Not 100% sure about just "staying in the current directory" - that feels like it would be unexpected behavior to me. If I run a command `ghost <cmd> --dir <some dir>` then I was intentionally wanting to run it in a specific directory, so having Ghost decide to stay in the current directory rather than error & exit seems surprising.

lib/bootstrap.js
```diff
@@ -145,7 +145,11 @@ const bootstrap = {
         // TODO: evaluate if `wrap` is actually needed?
         yargs.wrap(Math.min(100, yargs.terminalWidth()))
             .epilogue('For more information, see our docs at https://docs.ghost.org/v1/docs/ghost-cli')
-            .option('D', {
+            .option('d', {
+                alias: 'dir',
+                describe: 'Folder to run command in',
+                default: './'
```
acburdine: Not sure if we should have a default value - I think setting it to `type: 'string'` and then only executing the code in `command.js` when `argv.dir` is defined is maybe a better approach? Not sure what the performance implications of calling `process.chdir` are, but it might be better to not call it if it's not needed 😄

lib/command.js
vikaspotluri123: 1. Definitely!
2. Makes sense, will do

vikaspotluri123:
@acburdine ready for review! Any idea why the UI defaults test started failing all of a sudden?

acburdine:
Hmm, that's odd. You might try rebasing against master if you haven't already.

vikaspotluri123:
I just pulled from upstream and ran rebase to ensure the rebase actually happened, and it did 😕

vikaspotluri123:
I tried rebasing again to no avail - I'll see if I can figure out what's up tomorrow

vikaspotluri123:
I figured out the error - it has to do with the tests, but I'm not sure _why_ it's happening.

What's happening is a UI is being instantiated in the Command class as expected, but somehow (and this is what I don't understand) the options are being stored and being used in _other instances_ - for example, in the additional tests I added, I passed the `verbose: true` argument to the Command class, and if you look at the error in the build, it now reads `expected true to be false` instead of the original error (`expected undefined to be false`). Oddly enough passing `verbose: false` as a band-aid fix doesn't do anything

acburdine:
@vikaspotluri123 was able to find a fix for the tests, change [this line](https://github.com/vikaspotluri123/Ghost-CLI/blob/feat/global-dir/lib/ui/index.js#L38) to:

```js
this.options = Object.assign({}, defaultOptions, options);
```

what was happening was the default options were being overridden when the UI class was instantiated multiple times.

acburdine:
apologies for the delay! will get this looked at today.

PR #744, opened 2018-06-03
fix(run): local process manager forwards Ghost error message correctly

### LINKED ISSUE #742: IPC error handling
## This issue is a

* [x] Bug Report
* [ ] Feature Request

### Summary

There is something wrong when receiving an error from Ghost. I was testing with the local process manager only - not sure if this affects production as well. It's important to fix this error, because it's super hard for the user to know what's wrong if we ship [dynamic routing](https://github.com/TryGhost/Ghost/issues/9601), but the error is not forwarded correctly e.g. you change the routing file, but you received an error.

1. Ghost sends an IPC message e.g. `{started: false, error: 'a message'}`
2. The message is received [here](https://github.com/TryGhost/Ghost-CLI/blob/1.8.0/lib/commands/run.js#L71). It forwards only the message, no object.
3. The target process (which is in this case the local process manager) receives the error [here](https://github.com/TryGhost/Ghost-CLI/blob/1.8.0/lib/utils/local-process.js#L139). And here is the problem: this function tries to access `error.message`.

I guess we can simply change

> instance.process.error(message.error);

to

> instance.process.error({message:message.error, started: message.started});

1. Throw an error with the local process manager.
2. See that the error log in the CLI does not show the error message.

### Technical details (for a bug report)

* OS: Mac
* Node Version: 8
* Ghost-CLI Version: Latest
* Environment: development
* Command: ghost start

@acburdine Any opinion on that? Thanks!

kirrg001:
closes #742

- the target process manager has not received the correct error format

I have tested production - it's different. On production we do the port polling and if Ghost does not start (because it's not a child process where we can easily receive an IPC message), we print the journalctl command to get the information why Ghost has not started.

### DIFF
--- a/lib/utils/local-process.js
+++ b/lib/utils/local-process.js
@@ -60,7 +60,7 @@ Please ensure the content folder has correct permissions and try again.`));
             cp.on('message', (msg) => {
                 if (msg.error) {
                     fs.removeSync(path.join(cwd, PID_FILE));
-                    return reject(new errors.GhostError(msg.error));
+                    return reject(new errors.GhostError(msg.message));
                 }
 
                 /* istanbul ignore else */
--- a/test/unit/utils/local-process-spec.js
+++ b/test/unit/utils/local-process-spec.js
@@ -196,7 +196,7 @@ describe('Unit: Utils > local-process', function () {
                 done();
             });
 
-            cp.emit('message', {error: 'Test Error Message'});
+            cp.emit('message', {error: true, message: 'Test Error Message'});
         });
 
         it('writes pid to file, resolves on start message', function (done) {
--- a/lib/commands/run.js
+++ b/lib/commands/run.js
@@ -68,7 +68,7 @@ class RunCommand extends Command {
             }
 
             // NOTE: Backwards compatibility of https://github.com/TryGhost/Ghost/pull/9440
-            setTimeout(() => {instance.process.error(message.error);}, 1000);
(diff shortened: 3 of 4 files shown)


lib/utils/local-process.js
```diff
@@ -60,7 +60,7 @@ Please ensure the content folder has correct permissions and try again.`));
             cp.on('message', (msg) => {
                 if (msg.error) {
```
acburdine: this would need to be updated to be `msg.message` I think.

test/unit/utils/local-process-spec.js
```diff
@@ -196,7 +196,7 @@ describe('Unit: Utils > local-process', function () {
                 done();
             });
 
-            cp.emit('message', {error: 'Test Error Message'});
+            cp.emit('message', {error: true, message: 'Test Error Message'});
```
vikaspotluri123: @acburdine this would need to be updated as well, right?

lib/utils/local-process.js
kirrg001: Why? We forward `error: true`.

acburdine: ah, sorry. Misread a couple of things 🤦‍♂️

kirrg001: No problem :)

kirrg001:
@acburdine 🎊

acburdine:
Tested this locally and it works 👍

PR #749, opened 2018-06-11
fix(url): ensure trailing slash gets appended to url

### LINKED ISSUE #748: Admin interface location incorrect
### Summary

When running ghost, if a subdirectory is configured but no explicit admin url is configured the CLI outputs the admin's location incorrectly.

1. Create a local install (`ghost install local`)
2. Configure the url to be a subdirectory (`config set url http://localhost:2369/test` - change depending on the port that the CLI gave you)
3. Start ghost (`ghost start`)

Ghost-CLI will now output info about the Ghost install:

```
❯ ghost start
Found a development config but not a production config, running in development mode instead
✔ Validating config
✔ Checking memory availability
✔ Starting Ghost
You can access your publication at http://localhost:2369/test
Your admin interface is located at http://localhost:2369/ghost/
```

The admin URL here should be http://localhost:2369/test/ghost/

### Technical details (for a bug report)

* OS: macOS High Sierra
* Node Version: 8.10.0
* Ghost-CLI Version: 1.8.1
* Environment: development / local

vikaspotluri123:
closes #748 

The root of the problem stems from the fact we don't ensure trailing slashes when we set the URL. It's well-known that Ghost redirects URLs w/o a trailing slash to the trailing slash equivalent, so we might as well add the trailing slash here

Something that I would like to do but I'm not sure how to go about doing is add the trailing slash for existing installations. I know I can add a migration, but we also need to expect the URL will not be correct since it can be modified by running `ghost config set url https://my.site/blog`

### DIFF
--- a/lib/commands/start.js
+++ b/lib/commands/start.js
@@ -22,7 +22,6 @@ class StartCommand extends Command {
 
     run(argv) {
         const ProcessManager = require('../process-manager');
-        const url  = require('url');
         const chalk = require('chalk');
 
         const instance = this.system.getInstance();
@@ -64,7 +63,8 @@ class StartCommand extends Command {
                         const isInstall = process.argv[2] === 'install';
                         let adminUrl = instance.config.get('admin.url') || instance.config.get('url');
 
-                        adminUrl = url.resolve(adminUrl, '/ghost/');
+                        // Strip the trailing slash and add the admin path
+                        adminUrl = `${adminUrl.replace(/\/$/,'')}/ghost/`;
 
                         this.ui.log(`You can access your publication at ${chalk.cyan(instance.config.get('url'))}`, 'white');
 
--- a/test/unit/commands/start-spec.js
+++ b/test/unit/commands/start-spec.js
@@ -208,6 +208,26 @@ describe('Unit: Commands > Start', function () {
             });
         });
 
+        it('is normally loud', function () {
+            myInstance.config.get.withArgs('url').returns('https://my-amazing.site/blog');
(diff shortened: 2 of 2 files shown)


acburdine:
@vikaspotluri123 I was thinking we'd just fake the trailing slash when we build the admin url - would be a lot simpler.

vikaspotluri123:
Alright, that's fine 😄 Do we _have_ to use the url module in this case? We could just as easily just do `adminUrl += 'ghost/'` and it would do the same thing, with the added benefit of being more readable - if `url.resolve`'s second argument start's with a preceding `/`, it's considered an absolute path and the path from the first argument is removed

acburdine:
yeah - that would be fine too :)

vikaspotluri123:
Done! 👍

lib/commands/start.js
```diff
@@ -64,7 +63,8 @@ class StartCommand extends Command {
                         const isInstall = process.argv[2] === 'install';
                         let adminUrl = instance.config.get('admin.url') || instance.config.get('url');
 
-                        adminUrl = url.resolve(adminUrl, '/ghost/');
+                        // Strip the trailing slash and add the admin path
+                        adminUrl = adminUrl.replace(/\/$/,'') + '/ghost/';
```
acburdine: just to keep things consistent - can this be a template string?

vikaspotluri123: Yep, done 😃

vikaspotluri123:
Nope, I just forgot to get rid of it!

PR #752, opened 2018-06-14
chore(deps): upgrade non-breaking dependencies

vikaspotluri123:
The only dependencies that haven't been updated are rxjs because the error catching functionality isn't working w/ 6.x and validator because (I think) the newer URL validator doesn't work with local domains


acburdine:
probably more dep updates now, but will go ahead and merge this 😄

PR #755, opened 2018-07-13
feat(startup): improve startup time by lazily requiring ui deps

acburdine:
no issue
- dep: import-lazy@3.1.0
- add debug instrumentation to bin file
- lazily-import ui dependencies

Before these changes (key line is the "bootstrap loaded" line):
screen shot 2018-07-13 at 6 15 57 pm

After these changes:
screen shot 2018-07-13 at 6 16 55 pm

The biggest 2 offenders (found by instrumenting each require in the UI library, were Listr (which took about 1s) and inquirer (which also took about 1s). Lazily requiring each of these reduced the startup time fairly significantly.

### DIFF
--- a/lib/ui/index.js
+++ b/lib/ui/index.js
@@ -1,20 +1,19 @@
 'use strict';
-const ora       = require('ora');
-const omit      = require('lodash/omit');
-const chalk     = require('chalk');
-const execa     = require('execa');
-const Listr     = require('listr');
-const Table     = require('cli-table3');
-const Promise   = require('bluebird');
-const inquirer  = require('inquirer');
-const isObject  = require('lodash/isObject');
-const stripAnsi = require('strip-ansi');
-const ListrError = require('listr/lib/listr-error');
-const logSymbols = require('log-symbols');
-const isFunction = require('lodash/isFunction');
 
-const errors = require('../errors');
+const ora         = require('ora');
+const Promise     = require('bluebird');
+const isObject    = require('lodash/isObject');
+const logSymbols  = require('log-symbols');
+const isFunction  = require('lodash/isFunction');
+const lazyRequire = require('import-lazy')(require);
 const CLIRenderer = require('./renderer');
+
+const omit = lazyRequire('lodash/omit');
+const chalk = lazyRequire('chalk');
+const execa = lazyRequire('execa');
+const Listr = lazyRequire('listr');
+const Table = lazyRequire('cli-table3');
+const errors = lazyRequire('../errors');
 
 const defaultOptions = {
     stdin: process.stdin,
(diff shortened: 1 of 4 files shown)


lib/ui/index.js
```diff
@@ -1,21 +1,20 @@
...
+const ora         = require('ora');
+const Promise     = require('bluebird');
+const isObject    = require('lodash/isObject');
+const logSymbols  = require('log-symbols');
+const isFunction  = require('lodash/isFunction');
+const lazyRequire = require('import-lazy')(require);
 const CLIRenderer = require('./renderer');
 
+const omit = lazyRequire('lodash/omit');
+const chalk = lazyRequire('chalk');
```
vikaspotluri123: `log-symbols` requires chalk the normal way - is there any performance benefit by lazy loading it here?

lib/ui/index.js
```diff
@@ -44,16 +43,23 @@ class UI {
...
-        });
-
         CLIRenderer.ui = this;
     }
 
+    get inquirer() {
+        if (!this._inquirer) {
+            // Add custom prompt module that uses the
+            // specified streams
+            const inquirer = require('inquirer');
```
kirrg001: I am curios, why we not use lazyRequire for this case and move the dep on top? 🤔

lib/ui/index.js
kirrg001: It's also fully required [here](https://github.com/TryGhost/Ghost-CLI/blob/1.8.1/lib/commands/ls.js#L6) 🤔

kirrg001:
Loading dependencies when we need them makes sense to me 👍 like inquirer. 
But not sure about the lazy require lib 🤔 I get the idea, but if we only use it partially, it might be more confusing. e.g. you could just move the `listr` require into [this function](https://github.com/TryGhost/Ghost-CLI/blob/1.8.1/lib/ui/index.js#L166) and you will get almost the same performance results.

lib/ui/index.js
acburdine: was actually thinking about doing that - will move it.

acburdine:
@kirrg001 so the reason I didn't require the listr lib inside the `listr` function is that the listr function gets called multiple times - node _does_ use a require cache so it wouldn't be _that_ much of a performance penalty, but using the lazy import function means it only gets required once. I do see your point though about the confusing mix of lazy/regular imports though - I'll try and go through and normalize it.

acburdine:
So I was testing this a bit more, and one of the issues with the PR is that, while it _does_ improve startup time of certain commands like `ghost ls`,`ghost help`, etc, it actually hurts performance later on. If you run commands like `ghost install`, the startup time is the same, because it still needs to require Listr and inquirer (the 2 largest libs in terms of import cost).

Another approach that I also considered (cc @peterzimon on the UX of this) is to have a spinner appear as Ghost-CLI is loading/requiring files/setting up dependencies/etc, which then disappears after Ghost-CLI is ready to run. That might be a better approach simply because it makes it clearer to the user that something _is_ happening and that Ghost-CLI isn't just stuck.

Although, it _is_ only a couple of seconds really, so maybe it's not worth spending a ton of time on at the moment.

acburdine:
Gonna go ahead and close this for now - might revisit it later.

PR #765, opened 2018-07-25
Added handling for knex-migrator being executed in Ghost 2.0

kirrg001:
refs #759

This PR is **not finished**, please no code review. Feel free to add questions or concerns  👍 

The current port polling logic to detect if a blog is up or down on start/restart is too simple and the error handling was not good, because we weren't able to receive the reason why a blog didn't start up. Using the http port to communicate with simple messages between the CLI and Ghost is impossible, because it's the http port (!). Browsers use the same transport layer.

Ghost 2.0.0 has more requirements. It executes knex-migrator internally and brings the blog into maintenance mode. That means the server starts already, but the blog is not ready or fully started.

We just change how we communicate. We don't connect to Ghost, Ghost has to connect to the CLI.

- [x] use net server for communication
- [x] do final manual testing on ubuntu
- [x] ensure we haven't break the local process manager
- [x] fix tests
- [x] optimise error message for the websockets

This change is backwards compatible, because the CLI will support v1 and v2 (latest and previous major).

PR #768, opened 2018-07-28
feat: added support for named start

vikaspotluri123:
Adds functionality similar to the stop command - `ghost start {name}` now works as `ghost stop {name}` always has

I realized this wasn't a thing as I was testing #765, and given the Stop command already has this functionality, I figured it would be a 5 minute fix, but it turns out both the Start and Stop Command Unit tests aren't super dry, so I ended up going down the rabbit hole of drying up the tests. It might be easier to review commit by commit 😬 

I know this coincides w/ the release of 1.9.0; I don't think it will make much of a difference if this gets released before or afterwards.

### DIFF
--- a/lib/commands/doctor/index.js
+++ b/lib/commands/doctor/index.js
@@ -34,6 +34,8 @@ class DoctorCommand extends Command {
                 findValidInstall('doctor');
                 instance = this.system.getInstance();
                 instance.checkEnvironment();
+            } else {
+                instance = this.system.getInstance();
             }
 
             const context = {
--- a/lib/commands/doctor/index.js
+++ b/lib/commands/doctor/index.js
@@ -27,10 +27,7 @@ class DoctorCommand extends Command {
 
             if (
                 !argv.skipInstanceCheck &&
-                !(argv.categories &&
-                    argv.categories.length === 1 &&
-                    ['install', 'start'].includes(argv.categories[0])
-                )
+                !(argv.categories && argv.categories.length === 1 && argv.categories[0] === 'install')
             ) {
                 const findValidInstall = require('../../utils/find-valid-install');
 
--- a/lib/commands/doctor/index.js
+++ b/lib/commands/doctor/index.js
@@ -27,7 +27,10 @@ class DoctorCommand extends Command {
 
             if (
                 !argv.skipInstanceCheck &&
-                !(argv.categories && argv.categories.length === 1 && argv.categories[0] === 'install')
+                !(argv.categories &&
+                    argv.categories.length === 1 &&
+                    ['install', 'start'].includes(argv.categories[0])
+                )
             ) {
                 const findValidInstall = require('../../utils/find-valid-install');
 
--- a/lib/commands/start.js
+++ b/lib/commands/start.js
@@ -26,7 +26,12 @@ class StartCommand extends Command {
         const getInstance = require('../utils/get-instance');
 
         const runOptions = {quiet: argv.quiet};
-        const instance = getInstance(argv.name, this.system, 'start');
+        const instance = getInstance({
+            name: argv.name,
+            system: this.system,
+            command: 'start',
+            recurse: !argv.dir
+        });
 
         const isRunning = await instance.isRunning();
         if (isRunning) {
--- a/lib/commands/start.js
+++ b/lib/commands/start.js
@@ -23,8 +23,10 @@ class StartCommand extends Command {
     }
 
     async run(argv) {
-        const instance = this.system.getInstance();
+        const getInstance = require('../utils/get-instance');
+
         const runOptions = {quiet: argv.quiet};
+        const instance = getInstance(argv.name, this.system, 'start');
 
         const isRunning = await instance.isRunning();
         if (isRunning) {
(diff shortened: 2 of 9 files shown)


lib/commands/start.js
```diff
@@ -22,11 +22,18 @@ class StartCommand extends Command {
...
 
-        const instance = this.system.getInstance();
         const runOptions = {quiet: argv.quiet};
 
+        let instance;
+
+        try {
+            instance = getInstance(argv.name, this.system, 'start');
+        } catch (error) {
+            return Promise.reject(error);
```
acburdine: since we're not actually doing anything with the error here we can just make the above be:

```js
const instance = getInstance(argv.name, this.system, 'start');
```

any error generated by `getInstance` will get propagated fine, the `Promise.reject` wrapper is unnecessary.

lib/commands/stop.js
```diff
@@ -19,28 +19,23 @@ class StopCommand extends Command {
...
-        if (argv.name) {
-            if (!instance) {
-                return Promise.reject(new errors.SystemError(`Ghost instance '${argv.name}' does not exist`));
-            }
-
-            process.chdir(instance.dir);
+        try {
+            instance = getInstance(argv.name, this.system, 'stop');
+        } catch (error) {
+            return Promise.reject(error);
```
acburdine: same comment here as above

test/unit/commands/start-spec.js
```diff
@@ -1,32 +1,47 @@
...
 
+        it('uses getInstance to get proper instance', function () {
+            StartCommand = proxyquire(modulePath, {
+                '../utils/get-instance': sinon.stub().throws(new Error('YES_IT_DOES'))
+            });
+
+            const stop = new StartCommand();
+
+            try {
+                stop.run({}).then(() => {throw new Error('Expected promise rejection')});
```
vikaspotluri123: Any suggestion for avoiding the Promise error?

acburdine: for now, maybe wrap the `stop.run` call in a `Promise.resolve()`? that's how it's handled in the base run method I think

acburdine: What promise error are you talking about exactly? I'm not seeing one in the tests log (or when I run it locally)?

vikaspotluri123: Oops, I should have been clearer!

lib/utils/get-instance.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const checkValidInstall = require('./check-valid-install');
+const {SystemError} = require('../errors');
+
+function findInstance(instanceName, system, commandName) {
```
acburdine: should probably be `getInstance` for consistency with the filename

test/unit/commands/start-spec.js
acburdine: when would the invalid instance error get bypassed?

vikaspotluri123: Under normal circumstances, it shouldn't - I think I'm just overthinking it 🤕

acburdine: Ah ok - no problem 😄 These kind of issues should _hopefully_ be resolved once we can use async/await (i.e. when node 6 support is dropped).

vikaspotluri123:
@acburdine everything's been fixed now 😄

vikaspotluri123:
Rebased! 🙃

vikaspotluri123:
Couple things:

- Do we want to add parity between stop, start, and update? There is a [feature request](https://forum.ghost.org/t/ghost-update-multiple-sites/17565) for `update --all`, and we can add support for `start --all` as well.
- Should `get-instance` be used in the System Class?

lib/commands/doctor/index.js
```diff
@@ -27,7 +27,10 @@ class DoctorCommand extends Command {
 
             if (
                 !argv.skipInstanceCheck &&
-                !(argv.categories && argv.categories.length === 1 && argv.categories[0] === 'install')
+                !(argv.categories &&
```
acburdine, line 30: instead of making this change, could we pass `skipInstanceCheck: true` to the doctor command in start?

vikaspotluri123: Done

lib/commands/start.js
```diff
@@ -41,7 +48,7 @@ class StartCommand extends Command {
             ].join('\n'), 'yellow');
         }
 
-        await this.runCommand(DoctorCommand, {categories: ['start'], ...argv, quiet: true});
+        await this.runCommand(DoctorCommand, {categories: ['start'], ...argv, quiet: true, skipInstanceCheck: true});
```
vikaspotluri123, line 51: If we add support for named update, I think we can get rid of this `skipInstanceCheck` flag, since it will be true in all places where the `doctor` command is run. Plus, `start`, `stop`, and `update` will have similar command structures!

ErisDS:
Is there anything outstanding that prevents us from merging this PR? There may be more things we want to do, but if this change is valid and useful, let's get it in? 😄

vikaspotluri123:
Nothing on my side!

Issue #759, opened 2018-07-20
CLI support for Ghost 2.0


kirrg001:
[Ghost 2.0 is coming.](https://github.com/TryGhost/Ghost/issues/9742) 🤠

We will release a new **minor** CLI version before we release Ghost 2.0. This new CLI version will add support to upgrade your blog to Ghost 2.0. Furthermore the new CLI version will add support to stay on Ghost 1.0.

### Install/Update a Ghost 2.0 blog

If you execute `ghost update` and your are on 1.0, you will update to the next major. If you are on the next major (Ghost 2.0) and you execute `ghost update` you will update to the next Ghost 2.x release.

If you execute `ghost install` you will install a Ghost 2.0 blog.

### Install/Update a Ghost 1.0 blog

The CLI will add a new flag `--v1`. This flag will tell the CLI to either install or update a Ghost 1.0 blog (`ghost install --v1` or `ghost update --v1`). 

e.g. People who are still on LTS need an option to install a Ghost 1.0, because we disallow the jump from LTS to 2.0. We will only support the latest (2.0) and the previous (1.0) version. This is a new general concept we came up with. The same rule applies to [GScan](https://github.com/TryGhost/gscan/issues/116).

e.g. If we release a new 1.0 release after the Ghost 2.0 release, you can upgrade your blog to the latest 1.0 release with `ghost update --v1`.

### Guide to upgrade to Ghost 2.0

The CLI will add a proper support for the migration to Ghost 2.0

- print link to a new migration guide, which explains how the migration works
- scan your theme with GScan
- output the report from GScan
- print a link to the new demo draft post we will ship next week (a post about how the [new editor König](https://github.com/TryGhost/Ghost/issues/9724) works)
- the ability to continue/abort the major migration
- if the user continues
  - the CLI will download/install Ghost 2.0
  - Ghost 2.0 will use the same blog config
  - CLI will stop the 1.0 blog
  - CLI will start the 2.0 blog
  - Ghost 2.0 will bring the blog automatically into maintenance mode
  - Ghost 2.0 will then run the migration scripts
  - if successful, blog is out of maintenance mode
    - we use a net server communication (which is optional and needs to be configured from the outer unit)
  - if an error occurs during migrations, knex-migrator will auto rollback, the CLI must react and symlink the previous 1.0 version and start the blog
    - we have to send a socket + child event to receive the proper error message
- if the user aborts
  - the user stays on Ghost 1.0, nothing happens

[...]

PR #770, opened 2018-07-31
Avoid direct upgrade to v2

kirrg001:
refs #759

- extend version resolver
- detect major version jumps
- optimised code structure (use proper force option - this was hard to maintain/understand)

- [x] add unit tests
- [x] do manual test

### DIFF
--- a/lib/commands/update.js
+++ b/lib/commands/update.js
@@ -178,10 +178,9 @@ class UpdateCommand extends Command {
             return Promise.resolve(true);
         }
 
-        const updateVersion = force ? null : activeVersion;
         const resolveVersion = zip ?
-            () => require('../utils/version-from-zip')(zip, updateVersion) :
-            () => require('../utils/resolve-version')(version, updateVersion, v1);
+            () => require('../utils/version-from-zip')(zip, activeVersion, force) :
+            () => require('../utils/resolve-version')(version, activeVersion, v1, force);
 
         return resolveVersion().then((version) => {
             context.version = version;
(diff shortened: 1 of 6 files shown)


kirrg001:
@acburdine This PR requires the implementation of the `--v1` flag support.

vikaspotluri123:
Would we want to cache the results of the `yarn info` call? It seems wasteful to get all the versions of Ghost 2x 😛

kirrg001:
Waiting for a merge of #772. Then i'll finalise and optimise this PR.

kirrg001:
> Would we want to cache the results of the yarn info call? It seems wasteful to get all the versions of Ghost 2x 😛

Not sure i understand 🤔

kirrg001:
I see. Yeah we can do that 👍

kirrg001:
> Basically since versionResolver gets called twice (2x) there's no point in making the yarn call twice

I have changed the code - it was ugly to call the resolver twice. Cache is not needed anymore :)

kirrg001:
@vikaspotluri123 @acburdine This is ready for review. Would be cool if you could double check if i have missed a case.

kirrg001:
Okay i just discovered one case:

- You are on v2.
- You run `ghost update --force --v1`
- We should throw an error.
- Will add

kirrg001:
> Okay i just discovered one case:

Added

kirrg001:
I'll merge this into the 1.9 branch now. I'll open a new PR tomorrow to see all the changes from 1.9 and i'll add a QA list. If you have time helping with testing, let me know!

PR #771, opened 2018-07-31
Added support for major blog update 

kirrg001:
refs #759

## TODO

- [x] add tests
- [x] fix tests
- [x] full manual test
- [x] update UI
- [x] abord when fatal errors are detected
- [x] it could be that the demo post doesn't exist
- [x] source out this huge function to ...?


lib/commands/update.js
```diff
@@ -128,6 +141,66 @@ class UpdateCommand extends Command {
...
+                    this.ui.log(`Rule: ${warning.rule}`);
+                    this.ui.log(`File: ${map(warning.failures, 'ref').join(',')}`);
+                });
+            }
+
+            if (!gscanReport.results.error.length && !gscanReport.results.warning.length) {
+                this.ui.log('Theme is compatible.\n')
+            }
+
+            return this.ui.prompt({
```
vikaspotluri123: You should be able to use `this.ui.confirm` here :)

kirrg001: 👍

lib/utils/major-support.js
```diff
@@ -0,0 +1,58 @@
...
+
+            return gscan.check(path.resolve(themeFolder, activeTheme));
+        })
+        .then((report) => {
+            gscanReport = gscan.format(report);
+
+            return connection.raw('SELECT uuid FROM posts WHERE slug="v2-demo-post";')
+        })
+        .then((demoPost) => {
+            if (demoPost.length) {
```
kirrg001: Demo post could be deleted already. Test!

PR #772, opened 2018-08-02
feat(flag): add --v1 flag to install and update

acburdine:
refs #759
- adds --v1 flag to install and update commands
- add arg to resolveVersion util that limits versions to 1.x releases

TODO:
- [x] manual testing
- [x] fix unit tests

PR #773, opened 2018-08-02
Added confirmation/support for automatically rolling back to previous version

vikaspotluri123:
refs #759 

The wording won't work because I want to rebase on #772 and #771 before finalizing how the function will work

### DIFF
--- a/lib/commands/update.js
+++ b/lib/commands/update.js
@@ -114,7 +114,14 @@ class UpdateCommand extends Command {
                     return;
                 }
 
-                return this.ui.listr(tasks, context);
+                return this.ui.listr(tasks, context)
+                    .catch((error) => {
+                        if (error instanceof errors.GhostError && !context.rollback) {
+                            return this.rollbackFromFail(error, context.version, argv['auto-rollback']);
+                        }
+
+                        throw error;
+                    });
             });
         });
     }
@@ -195,6 +202,32 @@ class UpdateCommand extends Command {
         });
     }
 
+    rollbackFromFail(error, newVer, force = false) {
+        const oldVer = this.system.getInstance().cliConfig.get('previous-version');
+        const question = `Unable to upgrade Ghost from v${oldVer} to v${newVer}. Would you like to revert back to v${oldVer}?`;
+
+        this.ui.error(error, this.system);
+        this.ui.log('\n\n');
+
+        if (force) {
+            return this.run({
+                restart: true,
+                rollback: true
+            });
+        }
+
+        return this.ui.confirm(question, true).then(answer => {
+            if (!answer) {
+                return Promise.resolve();
+            }
+
+            return this.run({
+                restart: true,
+                rollback: true
+            });
+        });
+    }
+
     link(context) {
         const symlinkSync = require('symlink-or-copy').sync;
 
@@ -230,6 +263,11 @@ UpdateCommand.options = {
     },
     v1: {
         describe: 'Limit update to Ghost 1.x releases',
+        type: 'boolean',
+        default: false
+    },
+    'auto-rollback': {
+        description: '[--no-auto-rollback] Enable/Disable automatically rolling back Ghost if updating fails',
         type: 'boolean',
         default: false
     }
--- a/lib/tasks/migrator.js
+++ b/lib/tasks/migrator.js
@@ -9,7 +9,9 @@ const errorHandler = (context) => {
     const {config} = context.instance;
 
     return (error) => {
-        if (error.stderr && error.stderr.match(/CODE: ENOTFOUND/)) {
+        if (error.stderr && error.stderr.match(/No migrations available to rollback/)) {
+            return Promise.resolve();
+        } else if (error.stderr && error.stderr.match(/CODE: ENOTFOUND/)) {
             // Database not found
             error = new errors.ConfigError({
                 config: {
@@ -36,7 +38,7 @@ const errorHandler = (context) => {
             });
         } else {
             // only show suggestion on `ghost update`
-            error = new errors.ProcessError({
+            error = new errors.GhostError({
                 message: 'The database migration in Ghost encountered an error.',
                 stderr: error.stderr,
                 environment: context.instance.system.environment,
--- a/lib/utils/resolve-version.js
+++ b/lib/utils/resolve-version.js
@@ -29,7 +29,15 @@ module.exports = function resolveVersion(version, activeVersion, v1 = false, for
     }
 
     return yarn(['info', 'ghost', 'versions', '--json']).then((result) => {
-        let comparator = !force && activeVersion ? `>${activeVersion}` : MIN_RELEASE;
+        let comparator;
+
+        if (!force && activeVersion) {
+            comparator = `>${activeVersion}`;
+        } else if (force && activeVersion) {
+            comparator = `>=${activeVersion}`;
+        } else {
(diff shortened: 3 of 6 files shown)


lib/commands/update.js
```diff
@@ -179,6 +181,25 @@ class UpdateCommand extends Command {
         });
     }
 
+    rollbackFromFail(error, argv) {
+        const previousVersion = this.system.getInstance().cliConfig.get('previous-version');
+        this.ui.error(error);
+        const question = argv.version === '2.0.0' ?
```
kirrg001: Not sure we need to differentiate the text here 🤔 What was your thinking? :)

lib/commands/update.js
```diff
@@ -179,6 +181,25 @@ class UpdateCommand extends Command {
...
+            `Unable to upgrade Ghost - Would you like to revert back to ${previousVersion}?`;
+        const promise = argv['auto-rollback'] ? Promise.resolve(true) : this.ui.confirm(question, true);
+
+        return promise.then(answer => {
+            if (!answer) {
+                return Promise.resolve();
+            }
+
+            return this.run({
+                rollback: true
```
kirrg001: We can ensure a database rollback with https://github.com/TryGhost/knex-migrator/pull/110 and https://github.com/TryGhost/Ghost-CLI/pull/774 👍

lib/commands/update.js
```diff
@@ -179,6 +181,25 @@ class UpdateCommand extends Command {
         });
     }
 
+    rollbackFromFail(error, argv) {
+        const previousVersion = this.system.getInstance().cliConfig.get('previous-version');
+        this.ui.error(error);
```
kirrg001: What is the output format of 

> this.ui.error(error);

?

lib/commands/update.js
kirrg001: Will review this PR when we have merged #774 :)

lib/commands/update.js
vikaspotluri123: I'm not 100% sure 😬 I think the error would be a ListrError so it would be something like
```
One or more errors occurred.

1 Restarting Ghost

try running ghost doctor...
{docs help}
```

I figured it would be nice to use our native error recording functionality, but if it leaves the user confused that might not be the best idea 🤔

/cc @acburdine

lib/commands/update.js
vikaspotluri123: It was a mix of abstracting the utility of the function and a more explicit message about 2.0

I originally wrote the snippet based on the logic of only rolling back if upgrading to 2.x fails, so the error was explicit in saying `Ghost 2.0`. As I was creating the PR I realized this is applicable to all updates, not just to 2.0

With specifically saying 2.0 I thought it would be useful as a reminder since it's a major update. Now that I think about it, it might be useful to specify the attempted version regardless of the semver type

kirrg001: Yeah would be nice to just show the version which has failed.

e.g. `Unable to upgrade from X to Y. Would you like to revert to X?`

vikaspotluri123:
@kirrg001 updated the error message, still not sure how we should handle the error

lib/commands/update.js
kirrg001: Thanks 👍

vikaspotluri123:
Tests have been added!

kirrg001:
I have to test two cases:

A: You want to migrate and an error occurs during the migration.
B: You have successfully migrated, but want to rollback.

kirrg001:
> @vikaspotluri123 I'll take a look at this PR tomorrow :) Will also look at the error handling.

Tomorrow! 👻

lib/commands/update.js
```diff
@@ -195,6 +197,24 @@ class UpdateCommand extends Command {
         });
     }
 
+    rollbackFromFail(error, newVer, force = false) {
+        const oldVer = this.system.getInstance().cliConfig.get('previous-version');
+        const question = `Unable to upgrade Ghost from v${oldVer} to v${newVer}. Would you like to revert back to v${oldVer}?`;
+        const promise = force ? Promise.resolve(true) : this.ui.confirm(question, true);
+
+        this.ui.error(error);
```
kirrg001: This fails for me. `system` is missing as second argument.

lib/commands/update.js
```diff
@@ -195,6 +197,24 @@ class UpdateCommand extends Command {
         });
     }
 
+    rollbackFromFail(error, newVer, force = false) {
+        const oldVer = this.system.getInstance().cliConfig.get('previous-version');
+        const question = `Unable to upgrade Ghost from v${oldVer} to v${newVer}. Would you like to revert back to v${oldVer}?`;
+        const promise = force ? Promise.resolve(true) : this.ui.confirm(question, true);
```
kirrg001: We need to reorder the executions here, because the log output is broken. That is because we already trigger the confirm screen before we output the error. 

1. print error
1. if force, return this.run
2. run ui.confirm

lib/commands/update.js
```diff
@@ -114,7 +114,9 @@ class UpdateCommand extends Command {
                     return;
                 }
 
-                return this.ui.listr(tasks, context);
+                return this.ui.listr(tasks, context).catch(error =>
+                    this.rollbackFromFail(error, context.version, argv['auto-rollback'])
```
kirrg001: We have to differentiate between

1. errors which are migration/ghost errors
2. errors which happened in the CLI

kirrg001: ```if (error instanceof errors.GhostError)```

kirrg001: Also: we should not ask to rollback if you have executed `ghost update --rollback`.

lib/commands/update.js
```diff
@@ -195,6 +197,24 @@ class UpdateCommand extends Command {
         });
     }
 
+    rollbackFromFail(error, newVer, force = false) {
```
kirrg001: For v1 blogs we don't execute `knex-migrator rollback` yet. We should probably skip this function if you upgrade from 1.24 to 1.25 and it fails.

lib/commands/update.js
```diff
@@ -114,7 +114,9 @@ class UpdateCommand extends Command {
                     return;
                 }
 
-                return this.ui.listr(tasks, context);
```
kirrg001: Note to mysql: `ghost update --force` should update to 2.0 if you are on the latest version. This is currently not the case, it always installs the latest v1 release. Will open a PR soon.

kirrg001:
I'll push changes in a bit.

lib/commands/update.js
kirrg001: Fixed https://github.com/TryGhost/Ghost-CLI/pull/773/commits/435ad4d1e761e0715d15c0e8c50632691a08f961

lib/commands/update.js
```diff
@@ -195,6 +197,24 @@ class UpdateCommand extends Command {
...
+        const promise = force ? Promise.resolve(true) : this.ui.confirm(question, true);
+
+        this.ui.error(error);
+
+        return promise.then(answer => {
+            if (!answer) {
+                return Promise.resolve();
+            }
+
+            return this.run({
```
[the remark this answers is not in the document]
kirrg001: Added `restart:true`.

kirrg001:
I am doing a final test, then i'll merge this into 1.9.

PR #774, opened 2018-08-02
Executed database rollback on `ghost update --rollback`

### LINKED ISSUE #699: `ghost update --rollback` needs to trigger a database rollback
## This issue is a

* [x] Bug Report

Please fill out this template, deleting where appropriate:

### Summary

`ghost update --rollback` needs to trigger a database rollback.

1. You are on 1.21.7
2. You update to 1.22.1
3. Now you call `ghost update --rollback`
4. You are back on 1.21.7
5. But you have not rolled back the database.
6. You insert posts on 1.21.7, publish them
7. Now you call `ghost update` again
8. You are on 1.22, but the migration scripts are not re-triggered, because they ran already.
9. You will run into problems, you won't be able to serve the posts you've added in between, because multiple authors relationships were not created for those posts. Ghost could potentially ignore this error instead of throwing an error, but then we don't know that we are producing corrupt database states.

The migration scripts in `1.22` add a new table and migrate all posts to support multiple authors.

`knex-migrator rollback` has to be executed on the 1.22 installation folder.
Then you can switch back to 1.21.7.

* OS: any
* Node Version: any
* Ghost-CLI Version: latest
* Environment: any
* Command: any

### TODO

- [x] execute db rollback in the CLI at all (trigger knex-migrator)
- [x] extend knex-migrator's feature set (rollback to specific version) (https://github.com/TryGhost/knex-migrator/pull/110)

kirrg001:
closes #699

## TODO

- [x] manual test on mysql/sqlite


kirrg001:
@acburdine @vikaspotluri123 Ready to review/testing.

I'll try to extend the rollback feature in knex-migrator today/tomorrow. Currently we can only rollback to version you are on, but knex-migrator should rollback any version between the current and the previous version. 

We can merge this PR without the extension, because this logic triggers the db rollback at all.

kirrg001:
https://github.com/TryGhost/knex-migrator/pull/110

Will final test & release later or tomorrow morning.

lib/tasks/migrator.js
```diff
@@ -64,5 +46,53 @@ module.exports = function runMigrations(context) {
...
+};
+
+module.exports.rollback = function runRollback(context) {
+    const {dir} = context.instance;
+
+    const contentDir = path.join(dir, 'content');
+    const currentDir = path.join(dir, 'current');
+    let knexMigratorPromise;
+
+    const args = ['--force', '--version', context.instance.cliConfig.get('previous-version'), '--mgpath', currentDir];
```
kirrg001: The `--version` option requires the new version of knex-migrator in Ghost.
We have to add a bridge here (execute only for v2). A little annoying, but not sure how else to solve.
This is the disadvantage of not having knex-migrator as dependency in the CLI.

kirrg001:
It's fine if Ghost v1 updates to the latest knex-migrator version, but the CLI will only forward the `--version` flag on v2 blogs. Rolling back to a specific version will only apply to v2 blogs. 

This is getting important as soon as we push new features and you jump from e.g. 2.0.0 to 2.3.0 and you have to rollback, because something broke (rollback all migration scripts in 2.3, 2.2, 2.1).

kirrg001:
@vikaspotluri123 @acburdine Please let me know if you have time to review. Otherwise i'll just merge this into the 1.9 branch. Thanks!

lib/tasks/migrator.js
```diff
@@ -64,5 +46,60 @@ module.exports = function runMigrations(context) {
...
         return Promise.reject(error);
-    });
-}
+    }
+};
+
+module.exports.migrate = function runMigrations(context) {
+    const {dir} = context.instance;
+
+    const contentDir = path.join(dir, 'content');
```
vikaspotluri123: I don't think this was added in this PR, but shouldn't we programattically get this (i.e. `instance.config.get('paths.contentPath')`)

PR #777, opened 2018-08-08
Fix(migrations): allow existing content/settings folder

### LINKED ISSUE #776: Protect against existing settings folder
From https://forum.ghost.org/t/trying-to-upgrade-from-1-8-6/2318/13. Please read conversation.

I am not sure how this user came into the situation. Somehow he upgraded from an older Ghost version and Ghost-CLI complained that the `content/settings` folder exists already. I think we should just add a protection to our migration file, [see](https://github.com/TryGhost/Ghost-CLI/blob/1.8.1/lib/migrations.js).

Any help is welcome!

vikaspotluri123:
closes #776

In production sites, when the ghost user is used, the content/* folder needs to be owned by the Ghost user so the ghost application will not run into permission issues. In the migration, to make sure this happens, we run the `mkdir` command using sudo (i.e. `sudo -E -u ghost mkdir {path}`). When the Ghost user isn't needed, we use the `fs-extra` library to make sure the directory exists (specifically `ensureDirSync`). `fs.ensureDirSync` is equivilant to `mkdir -p` (not `mkdir`), so the behaviour differs. By adding the `p` flag to the `mkdir` command that's run when the Ghost user is used, the behaviour between the two cases are equalized, including support for a settings folder that already exists

### DIFF
--- a/lib/migrations.js
+++ b/lib/migrations.js
@@ -7,7 +7,7 @@ function ensureSettingsFolder(context) {
     const contentDir = context.instance.config.get('paths.contentPath');
 
     if (ghostUser.shouldUseGhostUser(contentDir)) {
-        return context.ui.sudo(`mkdir ${path.resolve(contentDir, 'settings')}`, {sudoArgs: '-E -u ghost'});
+        return context.ui.sudo(`mkdir -p ${path.resolve(contentDir, 'settings')}`, {sudoArgs: '-E -u ghost'});
     } else {
         const fs = require('fs-extra');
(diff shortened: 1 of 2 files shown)


acburdine:
Gonna go ahead and rebase this against the master branch - it's separate from the ghost 2.0 work so I think it's worth it to keep the eventual 1.9 PR a bit smaller :)

acburdine:
....and that promptly broke things lol. @vikaspotluri123 would you mind rebasing this branch against master?

vikaspotluri123:
Done!

PR #784, opened 2018-08-15
1.9: fix prerelease -> prerelease upgrades + dep updates

vikaspotluri123:
- allow upgrades from prerelease -> prerelease (i.e. Ghost 2.0.0-rc.1 -> Ghost 2.0.0-rc.2)
- Dependency upgrades


lib/utils/version-from-zip.js
```diff
@@ -69,7 +69,10 @@ module.exports = function versionFromZip(zipPath, currentVersion, force = false)
         })
         .catch((err) => {
             // CASE: you are on latest v1, but you want to manually migrate to v2 using a zip
-            if (err.message.match(/No valid versions found/) && semver.major(currentVersion) !== semver.major(pkg.version)) {
+            // CASE: you are on v2 prerelease, and want to upgrade to the next prerelease
+            if (err.message.match(/No valid versions found/) &&
+                (semver.major(currentVersion) !== semver.major(pkg.version) || semver.prerelease(pkg.version))
```
kirrg001: Logic looks correct. Thanks :)
Could you pls add one unit test case?

vikaspotluri123: Yep, oops 🙈

vikaspotluri123: fixed!

vikaspotluri123:
I'm not sure what changed in eslint to require only the bin folder to be globbed 😕 I tested that the other folders were being linted by adding a syntax error to a file, and everything else is being picked up properly 🤷‍♂️

kirrg001:
You have jumped to a major version of eslint?

vikaspotluri123:
Yes, since [the migration guide](https://eslint.org/docs/user-guide/migrating-to-5.0.0#migrating-to-v500) didn't seem to have any changes that would impact the CLI

PR #785, opened 2018-08-16
Separate URI Path and Domain

PhoenixPeca:
Missing slash breaks the link.

image

### DIFF
--- a/lib/tasks/major-update/ui.js
+++ b/lib/tasks/major-update/ui.js
@@ -22,7 +22,7 @@ module.exports = function ui(ctx) {
             ctx.ui.log(`${logSymbols.success} Your theme is compatible.\n`);
 
             if (demoPost && demoPost.uuid) {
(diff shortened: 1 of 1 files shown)


lib/tasks/major-update/ui.js
```diff
@@ -22,7 +22,7 @@ module.exports = function ui(ctx) {
             ctx.ui.log(`${logSymbols.success} Your theme is compatible.\n`);
 
             if (demoPost && demoPost.uuid) {
-                const demoLink = `${ctx.instance.config.get('url')}p/${demoPost.uuid}/`;
+                const demoLink = `${ctx.instance.config.get('url')}/p/${demoPost.uuid}/`;
```
vikaspotluri123: you probably also want to make sure you don't get a double trailing slash

```js
const demoLink = `${ctx.instance.config.get('url').replace(/\/$/,'')}/p/${demoPost.uuid}/`;
```

PhoenixPeca: Thanks! 😄

Issue #806, opened 2018-08-31
Introduce `--auto` flag for a simpler install


ErisDS:
We've been doing some user testing, and found that in certain contexts, the default behaviour of the CLI is asking the user to make too many decisions. The `--no-prompt` behaviour is better in most cases - it just does the sensible thing - but then you aren't prompted even for the most fundamental details.

The clear use case here is the Digital Ocean one-click installer. They configure all the hard stuff - mysql settings etc, but the _user_ needs to enter their desired domain and an email address for SSL.

We (@JohnONolan & me) have talked through the various subcommands of the install command and I think I have a solution that is straightforward, works for each individual subcommand and provides us with a bit more power/flexibility in the CLI to customise what a user sees.

The idea is to introduce a new `--auto` flag.

`ghost config --auto`

In `ghost config` the `--auto` flag switches to _only_ prompt if there isn't a non-default value already set. E.g. it should be possible to configure all the mysql details, and then call `ghost config` and only be prompted for a url.

⚠️Aside: I also think that `ghost config --url` should _only_ prompt for URL, rather than doing all prompts. Maybe this should be a separate issue, but I think there is some shared reworking of `ghost config` needed to solve both?

`ghost setup --auto`

For `ghost setup` the `--auto` flag switches to assuming that the user wants the standard setup, don't prompt for mysql user, nginx, ssl etc. This is almost identical to the `--no-prompt` behaviour of `ghost setup` (which is **amazing**). The only difference is that the prompt for the SSL email address needs to happen.

The `--auto` flag should also set `--start` to true.

⚠️Aside: Not sure if I'm misunderstanding, but there seems to be a weird flow whereby you are prompted if you want to create the `ghost` mysql user, but then the user is not root and so nothing actually happens?

`ghost install --auto --db=mysql ...`

For `ghost install` the `--auto` flag doesn't have any specific behaviour - it passes the flag to the subcommands.

With all of these changes in place, it is now possible for a tool like the one-click installer to generate an `install` command that is pre-filled with all the mysql config e.g. `ghost install --auto --db=mysql --dbhost=localhost --dbuser=ghost --dbpass=xxx --dbname=ghost_prod`, and the resulting behaviour will be that the user is only prompted for their url and their email address - two questions that most people are able to answer.

[...]

PR #807, opened 2018-08-31
feat: add --auto flag to make setup quicker

acburdine:
refs #806 

todo:
- [x] test w/more things
- [x] fix unit tests

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -168,6 +168,11 @@ const bootstrap = {
                 type: 'boolean',
                 default: true
             })
+            .option('auto', {
+                describe: 'Automatically run as much as possible',
+                type: 'boolean',
+                default: false
+            })
             .version(false)
             .parse(argv);
     }
--- a/lib/command.js
+++ b/lib/command.js
@@ -115,7 +115,8 @@ class Command {
         // Create CLI-wide UI & System instances
         const ui = new UI({
             verbose: argv.verbose,
-            allowPrompt: argv.prompt
+            allowPrompt: argv.prompt,
+            auto: argv.auto
         });
 
         // This needs to run before the installation check
--- a/lib/commands/config/index.js
+++ b/lib/commands/config/index.js
@@ -102,7 +102,7 @@ class ConfigCommand extends Command {
                 type: 'input',
                 name: 'url',
                 message: 'Enter your blog URL:',
-                default: this.instance.config.get('url', 'http://localhost:2368'),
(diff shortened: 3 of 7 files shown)


vikaspotluri123:
Finally got a chance to test 😄 

I created a clean VM (which means I can't set up SSL, but everything else works) and ran `ghost install`. I was only prompted for my URL, mysql credentials and sudo password (to create the ghost user).

Running it normally prompted for URL, mysql host, user pass, and db name, to create a mysql ghost user, to set up nginx, to set up ssl, to set up systemd and if I want to start ghost.

All this makes perfect sense 👌 

Running `ghost config --auto` asks me for the URL, MySQL host, user pass, and db - based on the spec:

> In ghost config the --auto flag switches to only prompt if there isn't a non-default value already set. E.g. it should be possible to configure all the mysql details, and then call ghost config and only be prompted for a url.

this doesn't seem to be happening - is this planned for a separate PR?

same question applies to `ghost setup --auto`

acburdine:
> same question applies to `ghost setup --auto`

`ghost setup` calls `ghost config` under the hood so really it's just a matter of improving/fixing the first thing. The reason `ghost config --auto` still prompts for mysql credentials is that there _aren't_ defaults for those values,  you have to provide those manually. I'm working on a solution that should improve this usage, though.

ErisDS:
Just given this a quick test. The behaviour of `ghost config --auto` is as expected 🎉 

However, `ghost setup --auto` is still prompting me for each stage, rather than assuming the default like `--no-prompt` does.

acburdine:
@ErisDS I ran:

- `ghost install --no-setup`
- `ghost setup --auto`

I was, as expected, prompted for the url & mysql credentials, but I wasn't prompted for the setup stages (nginx, ssl, etc.). What series of commands did you run to get the behavior you were seeing?

ErisDS:
I'm testing the main use case, which is a pre-filled install command. E.g. `ghost install --auto --db=mysql --dbhost=localhost --dbname=ghost_2 --dbuser=ghost etc` 

All of the config for the DB is provided.

acburdine:
@ErisDS I ran that exact command (minus the `etc` part), and was still not able to reproduce. I was only prompted for the blog url and the ssl generation 😞.

ErisDS:
I'm running it on a DO droplet e.g. not a local install - in case that helps

ErisDS:
Managed to figure out what was wrong - this is good to go now.

PR #813, opened 2018-09-05
✨ Updated URL handling to perform basic formatting

### LINKED ISSUE #812: Make it easier to enter a URL during `ghost config`
At present we expect a fully formatted URL to be entered during `ghost config` including the protocol. However, we don't _really_ make that clear in all cases and it's not really required. It's perfectly reasonable for me to enter `mysite.com` and expect that the CLI would add the https part for me along with setting up SSL.

Instead, what happens currently is the CLI complains the URL isn't valid 😞 

I think we should add protocols to the URL according to the following rules:

- if the url entered is localhost, or an ip, add `http`
- if the url entered is a domain, add `https`

Note: we are moving as fast as we can towards only supporting http for development - all Ghost blogs should be run fully HTTPS in production.

ErisDS:
Part of a minor UX improvement project, we're trying to make it as easy as possible for people to get Ghost running using Ghost CLI and particularly with the DigitalOcean 1-click installer. 

closes #812

- When entering a URL, add a protocol if one is missing
- Follow the rules laid out in #812, so that we use https by default where possible
- Also handle domains entered with // at the start


acburdine:
@ErisDS I tested this out - looks like the fixURL stuff only works when you run `ghost install --url <my-url>`. If you run and get prompted for a URL, the same rules unfortunately don't apply.

However, if you add the fixURL function as a `filter` option [here](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/config/index.js#L106) it fixes the issue.

acburdine:
@ErisDS since it involves prompting, I don't unfortunately really have a test I can point at - all I can say is I tested it locally (including putting the fixURL function as a `filter` option to prompt) and I got it to work 🤷‍♂️

ErisDS:
Ok - it's just I had it as a filter the first time, and I couldn't get the tests to pass, hence why it's now a transform 😬

acburdine:
For clarification - it would _only_ need to be a `filter` in the getConfigPrompts section of the config command, not in the advanced options file.

ErisDS:
I've done a second pass, moved my domain functionality into utils along with the existing validator. This is partially because I have a second use case for `isCustomDomain` coming to another PR near  you.

ErisDS:
Just added one more commit, making the admin URL and url validation consistent. I think it was possibly different because of the `/ghost` path validation, but neither should contain it really?

No tests fail.

PR #815, opened 2018-09-06
🐛 Ensure all files are removed during uninstall

### LINKED ISSUE #811: `ghost uninstall` sometimes can't delete all content in directory
### Summary

If you get an issue where something gets created in your ghost install directory as root (such as on SSL verification), and then you try to run `ghost uninstall`, you'll get an error stating that it can't delete the directory (because it's owned by root)

### Steps to Reproduce

1. `ghost install`
2. SSL setup fails
3. `ghost uninstall`
4. Error, can't delete `system/files/.well-known`

This is because we're doing a simple `fs.remove` on the ghost install folder. Might be worth doing a `rm -rf` instead.

acburdine:
closes #811
- adds another task to unistall that removes root-owned files in the install directory

### DIFF
--- a/lib/commands/uninstall.js
+++ b/lib/commands/uninstall.js
@@ -2,21 +2,21 @@
 const Command = require('../command');
 
 class UninstallCommand extends Command {
-    run(argv) {
+    run({force}) {
         const fs = require('fs-extra');
         const path = require('path');
 
         const StopCommand = require('./stop');
-        const ghostUser = require('../utils/use-ghost-user');
+        const {shouldUseGhostUser} = require('../utils/use-ghost-user');
 
-        if (!argv.force) {
+        if (!force) {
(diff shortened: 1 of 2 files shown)


acburdine:
@ErisDS @kirrg001 could I get a quick review on this?

ErisDS:
How come the .well-known file gets created as root? Can we fix that?

kirrg001:
> How come the .well-known file gets created as root?

I wanted to ask that too.

Furthermore, how can i reproduce this without creating the .well-known file manually?

acburdine:
The `.well-known` directory gets created as root because we run `letsencrypt` with sudo (which is required to get the certs installed in `/etc/letsencrypt`)

acburdine:
@kirrg001 you can reproduce by chowning one of the files in `<ghost dir>/system` to root, then trying `ghost uninstall`.

PR #817, opened 2018-09-06
chore(deps): upgrade yargs to 12.0.1

vikaspotluri123:
- includes lockfile regeneration
- was held back for manual testing

The major breaking changes don't seem to impact the CLI, although I haven't tested a full `ghost install` with arguments

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -136,8 +136,8 @@ acorn@^3.0.4:
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-3.3.0.tgz#45e37fb39e8da3f25baee3ff5369e2bb5f22017a"
 
 acorn@^5.0.3, acorn@^5.5.0, acorn@^5.6.0:
-  version "5.7.1"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -420,7 +432,7 @@ buffer-from@^1.0.0:
 
 buffer@^3.0.1:
   version "3.6.0"
-  resolved "https://registry.yarnpkg.com/buffer/-/buffer-3.6.0.tgz#a72c936f77b96bf52f5f7e7b467180628551defb"
+  resolved "http://registry.npmjs.org/buffer/-/buffer-3.6.0.tgz#a72c936f77b96bf52f5f7e7b467180628551defb"
```
acburdine: @vikaspotluri123 do you know why yarn would switch the links to the non-https npm registry as opposed to the https yarn registry? (maybe your `.npmrc` is pointed to that?)

Would prefer all of the links remain pointed to https if possible 😄

vikaspotluri123: I'm not sure what's going on, I cleared my cache and tried again but it still pulled in some http links. Since all of them pointed to NPM registry I manually updated them with sed 😕

PR #818, opened 2018-09-06
fix(update): maintain instance state

vikaspotluri123:
no issue
running `ghost update` will start ghost no matter what, even if the instance is stopped

- add an optional `--[no]-restart` argument to explicitly define restart behavior
- by default, restart only if the instance is running

### DIFF
--- a/lib/commands/update.js
+++ b/lib/commands/update.js
@@ -58,6 +58,11 @@ class UpdateCommand extends Command {
         return instance.running().then((isRunning) => {
             if (isRunning) {
                 instance.loadRunningEnvironment(true);
+
+                // argv.restart will only be undefined if it wasn't passed in
+                if (argv.restart === undefined) {
+                    argv.restart = true;
+                }
             }
 
             instance.checkEnvironment();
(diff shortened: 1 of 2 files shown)


lib/commands/update.js
```diff
@@ -58,6 +58,11 @@ class UpdateCommand extends Command {
         return instance.running().then((isRunning) => {
             if (isRunning) {
                 instance.loadRunningEnvironment(true);
+
+                // argv.restart will only be undefined if it wasn't passed in
+                if (argv.restart === undefined) {
```
acburdine: might be better (since we've upgraded to yargs 12 which enables this), to use the `in` operator:
```suggestion
                if (!('restart' in argv)) {
```
imo it's a _little_ cleaner

lib/commands/update.js
```diff
@@ -107,7 +112,7 @@ class UpdateCommand extends Command {
                 }
             }, {
                 title: 'Restarting Ghost',
-                skip: () => !argv.restart,
+                enabled: () => argv.restart === true,
```
acburdine: don't necessarily need the === true here, esp. since we have the boolean type in the option config

vikaspotluri123: Done!

Issue #800, opened 2018-08-24
Support discovering instances in parent directories


vikaspotluri123:
### Summary

Similar to [git](https://git-scm.com/docs/git/1.7.6#git-emGITDISCOVERYACROSSFILESYSTEMem), it would be nice if the CLI was able to traverse up the path to discover a ghost instance - for example, if I'm developing a theme and I'm in my theme folder (`{ghost}/content/themes/my-awesome-theme`) it would be nice if I could run `ghost restart` in that folder and the CLI determine I'm trying to restart the instance with path `{ghost}`

vikaspotluri123:
#820 implemented that, so closing :smile:

PR #820, opened 2018-09-06
feat(instance): traverse up directory tree to find instance

vikaspotluri123:
refs #800 

- Try to go up the directory tree to find a valid Ghost instance
- Can be disabled with global config ('traverse_dirs')
  - We need to determine if there's a better way of doing this, or how global config should work in general
  - I was thinking of adding a migration which prompts the user, but that feels overkill
- I want to get a code review before I write the tests 😄

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -144,10 +144,8 @@ class Command {
         if (!this.global) {
             const findValidInstall = require('./utils/find-valid-install');
 
-            // CASE: directory was NOT specified (`-d /path/to/instance`)
-            // CASE: directory traversal is explicitly disabled
-            const shouldTraverse = !argv.dir && system.globalConfig.get('traverse_dirs', true);
-            findValidInstall(commandName, shouldTraverse);
+            // NOTE: we disable recursive searching when the cwd is supplied
+            findValidInstall(commandName, !argv.dir);
         }
 
         if (!this.allowRoot) {
--- a/lib/command.js
+++ b/lib/command.js
@@ -139,10 +139,15 @@ class Command {
             debug('Finished updating directory');
         }
 
-        if (!this.global) {
-            const checkValidInstall = require('./utils/check-valid-install');
+        const system = new System(ui, extensions);
 
-            checkValidInstall(commandName);
+        if (!this.global) {
+            const findValidInstall = require('./utils/find-valid-install');
+
+            // CASE: directory was NOT specified (`-d /path/to/instance`)
+            // CASE: directory traversal is explicitly disabled
+            const shouldTraverse = !argv.dir && system.globalConfig.get('traverse_dirs', true);
+            findValidInstall(commandName, shouldTraverse);
         }
 
         if (!this.allowRoot) {
@@ -150,8 +155,6 @@ class Command {
             // Check if user is trying to install as `root`
             checkRootUser(commandName);
         }
-
-        const system = new System(ui, extensions);
 
         // Set the initial environment based on args or NODE_ENV
         system.setEnvironment(argv.development || process.env.NODE_ENV === 'development', true);
--- a/lib/command.js
+++ b/lib/command.js
@@ -139,8 +139,6 @@ class Command {
             debug('Finished updating directory');
         }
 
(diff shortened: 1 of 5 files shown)


lib/utils/check-valid-install.js
```diff
@@ -21,33 +22,47 @@ function checkValidInstall(name) {
...
     }
 
     /*
-     * We assume it's a Ghost development install if 3 things are true:
-     * 1) package.json exists
-     * 2) package.json "name" field is "ghost"
-     * 3) Gruntfile.js exists
-     */
+    * We assume it's a Ghost development install if 3 things are true:
+    * 1) package.json exists
```
vikaspotluri123: This wasn't supposed to happen 🙅‍♂️

lib/command.js
```diff
@@ -141,10 +141,15 @@ class Command {
...
+        const system = new System(ui, extensions);
+
         if (!this.global) {
-            const checkValidInstall = require('./utils/check-valid-install');
+            const findValidInstall = require('./utils/find-valid-install');
 
-            checkValidInstall(commandName);
+            // CASE: directory was NOT specified (`-d /path/to/instance`)
+            // CASE: directory traversal is explicitly disabled
+            const shouldTraverse = !argv.dir && system.globalConfig.get('traverse_dirs', true);
```
acburdine: do we really need to make this a configurable setting? I guess there _could_ be privacy concerns, but I feel like people concerned about that level of privacy would also maybe chroot the user they run `ghost` commands with or change the read permissions to prevent the user from reading it.

Without a way to easily set this property (i.e. not opening the global config and editing it), I feel like adding in configurability isn't a huge need.

vikaspotluri123: It's been a while, but I think I added it because git supports something like that, but I agree, we could definitely get rid of thsi

lib/utils/check-valid-install.js
```diff
@@ -21,33 +22,47 @@ function checkValidInstall(name) {
 If you are trying to upgrade Ghost LTS to 1.0.0 please see ${chalk.blue.underline('https://docs.ghost.org/v1/docs/migrating-to-ghost-1-0-0')}.
 Otherwise, please run \`ghost ${name}\` again within a valid Ghost installation.`);
 
-        process.exit(1);
+        if (exit) {
+            process.exit(1);
+        }
+
+        throw new Error('UNSUPPORTED');
```
acburdine: would it make more sense to maybe return `true/false/null` instead of throwing errors? might clean up the find-valid-install code that way

vikaspotluri123: I think I've done this now 😄

lib/utils/find-valid-install.js
```diff
@@ -0,0 +1,38 @@
...
+        if (isValidInstall === false) {
+            process.exit(1);
+        }
+
+        if (!recursive) {
+            return die(name);
+        }
+
+        debug(`Going up...`);
+        process.chdir('..');
```
acburdine, line 32: apologies if I mentioned this in a comment before 😅 but would it make sense to, instead of `chdir`ing each time, to maintain `dir` string throughout the while loop, and just update it for each iteration where it doesn't find a directory?

i.e.
```js
let dir = process.cwd();
while(!isRoot(dir)) {
  const isValidInstall = checkValidInstall(name, dir);

  // checks
  debug('Going up...');
  dir = dirname(dir);
}
```
then do a chdir at the end? might be more performant

lib/utils/find-valid-install.js
```diff
@@ -0,0 +1,38 @@
...
+        if (isValidInstall) {
+            return;
+        }
+
+        if (isValidInstall === false) {
+            process.exit(1);
+        }
+
+        if (!recursive) {
+            return die(name);
```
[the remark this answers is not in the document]
vikaspotluri123: Yep

lib/utils/find-valid-install.js
vikaspotluri123: Taking a look

PR #822, opened 2018-09-07
UI Tweaks

ErisDS:
This PR builds on top of #813 & focuses on messaging.

I think I could fiddle with the output from CLI forever, but for now there are 5 changes:

1. reduced the sudo output to just add `+ sudo` before the command, consistent with the output from npm.
2. reworded the email prompt for SSL
3. removed the word URL from as many places as possible, when talking to users we should use "domain"
4. a whole bunch of work to the messages output at the end of a start/install to make this super clear
5. changed many messages to make them clearer

Bottom line with language in the CLI is that it should be direct and clear, and that clarity is way more important that politeness.

### DIFF
--- a/lib/commands/start.js
+++ b/lib/commands/start.js
@@ -1,6 +1,7 @@
 'use strict';
 const Command = require('../command');
 const DoctorCommand = require('./doctor');
+const urlUtils = require('../utils/url');
 
 class StartCommand extends Command {
     static configureOptions(commandName, yargs, extensions) {
@@ -22,14 +23,13 @@ class StartCommand extends Command {
 
     run(argv) {
         const ProcessManager = require('../process-manager');
-        const chalk = require('chalk');
 
         const instance = this.system.getInstance();
         const runOptions = {quiet: argv.quiet};
 
         return instance.running().then((isRunning) => {
             if (isRunning) {
-                this.ui.log('Ghost is already running! Run `ghost ls` for more information', 'green');
+                this.ui.log('Ghost is already running! For more information, run', 'ghost ls', 'green', 'cmd', true);
                 return Promise.resolve();
             }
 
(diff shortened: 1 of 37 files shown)


lib/commands/start.js
```diff
@@ -62,22 +62,27 @@ class StartCommand extends Command {
...
-                        this.ui.log(`You can access your publication at ${chalk.cyan(instance.config.get('url'))}`, 'white');
+                        // Show a warning about direct mail - but in grey, it's not the most important message here
+                        if (isInstall && instance.config.get('mail.transport') === 'Direct') {
+                            this.ui.log('\nGhost uses direct mail by default. To set up an alternative email method read our docs at https://ghost.org/mail', 'gray');
+                        }
+
+                        if (isInstall && instance.config.get('process') !== 'local' && !urlUtils.isCustomDomain(instance.config.get('url'))) {
+                            // Special message for when a production install is completed with an ip or localhost instead of a domain
+                            this.ui.log('\n------------------------------------------------------------------------------', 'white');
+                            this.ui.log('To add a custom domain and secure your site, run', 'ghost setup ssl --url example.com', 'white', 'cmd');
```
ErisDS: Hmmm @acburdine I just tested, and this command doesn't work

`ghost setup ssl --url https://myblogurl.com`

Results in

> SSL certs cannot be generated for IP addresses, skipping
ℹ Setting up SSL [skipped]

Tried with and without the equals sign.

ErisDS: To get it to work I had to do:

`ghost config url https://myblogurl.com`

`ghost setup nginx ssl`

ghost setup ssl on its own said:

> Nginx config file does not exist, skipping SSL setup
ℹ Setting up SSL [skipped]

acburdine: right...so when you run `ghost setup` with individual steps it bypasses the "configuration" part of things, and just runs the individual steps.

acburdine: should be able to fix that with the setup refactor I think

PR #829, opened 2018-09-26
chore: ensure docs links point to non-versioned docs

### LINKED ISSUE #828: Fix hardcoded docs links
A lot of the links in the CLI codebase are pointing at the old Ghost v1 docs ([search link](https://github.com/TryGhost/Ghost-CLI/search?q=v1%2Fdocs&unscoped_q=v1%2Fdocs)).

Where appropriate (meaning there might be places where links to the v1 docs _should_ remain), these should be updated with links to the non-v1 docs (e.g. just `docs.ghost.org/docs/<doc-page>`).

acburdine:
closes #828
- replace /v1 docs links with their non-versioned counterparts
- correct links pointing to the wrong section of docs

### DIFF
--- a/.github/ISSUE_TEMPLATE/--anything-else.md
+++ b/.github/ISSUE_TEMPLATE/--anything-else.md
@@ -14,5 +14,5 @@ Alternatively, check out these resources below. Thanks! 😁.
 - [Ghost-CLI docs](https://docs.ghost.org/docs/ghost-cli)
 - [Forum Support](https://forum.ghost.org/c/help)
 - [Ideas](https://forum.ghost.org/c/Ideas)
-- [Contributing Guide](https://docs.ghost.org/v1/docs/contributing)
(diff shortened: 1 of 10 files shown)


kirrg001:
Cool 👍 

Doesn't it depend on the context your blog is in? 🤔 
e.g. if you run a command on an 1.x blog, should we output 1.x docs links?

acburdine:
@kirrg001 I looked for cases where there would have been differences between 2.0 docs and 1.0 docs that would be contextually important. The only case I could think of was during database migrations, and the docs currently output there go to the troubleshooting docs, which are the same between 1.x and 2.x.

From a code standpoint too, I'm not 100% on outputting different docs version depending on the version of the blog, that seems like a lot of extra work that (depending on what the new docs structure ends up being) might end up being changed down the line.

kirrg001:
Ok 👍 I thought there are more Ghost docs links. Then it's fine for now :)

Issue #588, opened 2018-01-16
Improve setup command UX -> run as many steps as possible


acburdine:
One of the pain points that we've seen experienced by several people is the fact that the `setup` command currently exits if any one of it's steps fails.

Using the official system stack with all steps enabled, here's the list of steps that get run in order:
- configure
- setup linux user
- setup nginx
- setup ssl
- setup mysql user
- setup systemd
- run database migrations
- start Ghost

Currently, if, say, the SSL step fails (which it often does for people given that their DNS isn't correctly set up yet), _all_ of the steps after it are currently skipped, with no indication to the user that the setup command just _didn't_ run any of the other steps.

The setup command should be updated so that it runs all of the steps that it can, and if any failed after all the steps are done, print out a report saying something like (not final wording just an idea):

```
The following setup steps failed: ssl, systemd.
Fix any issues then run `ghost setup ssl systemd`
```

Setup steps already have the concept of "dependencies", e.g. if nginx setup is skipped, ssl setup will also be skipped. We can utilize this concept as well for errors - so if the nginx step errors the SSL step will skip.

acburdine:
This should theoretically be fixed as of the refactored setup command in Ghost-CLI 1.10.0. If it is found to not be fixed, we can reopen it 😄

PR #831, opened 2018-09-30
Refactor Setup Command (take 2)

### LINKED ISSUE #594: Running `ghost setup` should be (more) idempotent
A user reported a case where they ended up with two systemd service files, one called `ghost_<ip address of blog>` and one called `ghost_<url of blog>`. Not for sure on how this happened, but based on the current code I have a theory:

- User runs `ghost setup`, gives ip address as url
- After a while, user runs `ghost setup` again, but gives the actual domain name this time

In this case, `ghost setup` _could_ potentially set up two different systemd service files, because the ghost instance "name" is generated by using the url (see [here](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/setup.js#L109)). Quick fix would probably be to skip the setup step that sets the instance name if the name is already set....however it's probably worth investigating further to see if there are more things in the setup command that can be fixed to make it idempotent.

If I find more things, I will add them here:

- [ ] Ensure instance name doesn't get reset if it already exists

acburdine:
closes #594, #588 

todo:
- [ ] manual testing with all the setup steps
- [x] fix unit tests

### DIFF
--- a/extensions/mysql/index.js
+++ b/extensions/mysql/index.js
@@ -3,30 +3,28 @@
 const Promise = require('bluebird');
 const mysql = require('mysql');
 const omit = require('lodash/omit');
-const cli = require('../../lib');
 const generator = require('generate-password');
+const {Extension, errors} = require('../../lib');
 
 const localhostAliases = ['localhost', '127.0.0.1'];
+const {ConfigError, CliError} = errors;
 
-class MySQLExtension extends cli.Extension {
-    setup(cmd, argv) {
-        // Case 1: ghost install local OR ghost setup --local
-        // Case 2: ghost install --db sqlite3
-        // Skip in both cases
-        if (argv.local || argv.db === 'sqlite3') {
-            return;
-        }
-
-        cmd.addStage('mysql', this.setupMySQL.bind(this), [], '"ghost" mysql user');
+class MySQLExtension extends Extension {
+    setup() {
+        return [{
+            id: 'mysql',
+            name: '"ghost" mysql user',
+            task: (...args) => this.setupMySQL(...args),
+            // Case 1: ghost install local OR ghost setup --local
+            // Case 2: ghost install --db sqlite3
+            // Skip in both cases
+            enabled: ({argv}) => !(argv.local || argv.db === 'sqlite3'),
+            skip: ({instance}) => instance.config.get('database.connection.user') !== 'root'
+        }];
     }
 
-    setupMySQL(argv, ctx, task) {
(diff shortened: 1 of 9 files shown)


acburdine:
@ErisDS @kirrg001 @vikaspotluri123 there are still some aforementioned todos, but would appreciate an initial code review as I _think_ I've gotten most of the actual refactoring done.

lib/commands/setup.js
```diff
@@ -21,180 +21,144 @@ class SetupCommand extends Command {
...
 
-        if (argv.stages && argv.stages.length) {
-            const instance = this.system.getInstance();
+        const instance = this.system.getInstance();
+
+        // This is used so we can denote built-in setup steps
+        // and disable the "do you wish to set up x?" prompt
+        // We use symbols so extension-defined tasks can't
+        // mark themselves as internal and skip the prompt
+        const internal = Symbol('internal setup step');
```
acburdine: as far as I can tell this is what symbols are _supposed_ to be used for, and it seemed like a good solution.

extensions/nginx/index.js
```diff
@@ -25,23 +25,43 @@ class NginxExtension extends Extension {
...
+            name: 'SSL',
+            enabled,
+            task: (...args) => this.setupSSL(...args),
+            skip: ({tasks, ui}) => {
+                if (tasks.nginx.isSkipped()) {
+                    ui.log('Nginx setup task was skipped, skipping SSL setup', 'yellow');
+                    return true;
+                }
+
+                if (tasks.nginx.hasFailed()) {
```
acburdine: these two status checks could maybe be combined, seemed more prudent to leave the messaging different per case though.

extensions/mysql/index.js
```diff
@@ -53,7 +51,8 @@ class MySQLExtension extends cli.Extension {
 
         return Promise.fromCallback(cb => this.connection.connect(cb)).catch((error) => {
             if (error.code === 'ECONNREFUSED') {
-                return Promise.reject(new cli.errors.ConfigError({
+                return Promise.reject(new ConfigError({
+
```
vikaspotluri123: Not really pertinent, but is this extra line needed?

acburdine: Not sure how that got added 🙃

acburdine:
...aaaand it's done 🎉 

just needs manual testing & some commit squashing

acburdine:
Note: I split off some of the commits in this PR into #840 -> it'll shrink this PR a little bit.

PR #836, opened 2018-10-02
Add x-xss-protection Header and Hide Powered by Express

mskian:
1 - Add x-xss-protection Security Header
2 - Hide `X-Powered-By: Express`


acburdine:
@mskian I'm not _against_ this change necessarily, but before I approve this change I would appreciate some links to documentation as to why these changes are necessary/a good idea to apply to _all_ users of Ghost-CLI.

Also, the change as it stands now would only affect future users of the CLI, not current users, so you'd need to write a migration as well (see [here](https://github.com/TryGhost/Ghost-CLI/blob/master/extensions/nginx/index.js#L20) for an example) so that current users would get upgraded to use these changes as well.

acburdine:
@mskian Going to close this for now - if you can provide a response to my points above then we can reopen this 😄

Issue #787, opened 2018-08-18
Bootstrap port detection needs to be moved into the core


vikaspotluri123:
Currently, Process Manager extensions using the `ensureStart` method in CLI@1.8.x won't work with CLI@1.9.0 in Ghost 2.0. This is because the CLI uses socket creation instead of port polling for Ghost > 2.0.0.

The reason this is an issue is because port detection + config updates are done in the systemd extension, rather than in the core. We need to move this logic into the core, and if needed add logic to determine if a given process manager needs this functionality

acburdine:
I think I can work this into the changes I'm making as part of #736

PR #837, opened 2018-10-06
[Enhancement] Delete bootstrap-socket once ghost is stopped

### LINKED ISSUE #805: Remove bootstrap socket config after Ghost was started
### Summary

With CLI 1.9 we have invented a new mechanism for systemd to figure out if Ghost is running or not.

CLI writes into the blog's config file, but doesn't clean up it afterwards (`bootstrap-socket` options)

If you then run `ghost run` the bootstrap socket options still exist and Ghost tries to connect to the socket. That is all fine, Ghost is just trying and stops after 3 tries, but it's very confusing.

If we remove the config, the CLI will add it next time ghost get's started/restarted.

### Steps to Reproduce

1. Install Ghost on production using systemd
2. Start your blog
3. Stop your blog
4. run `ghost run`

Ghost-CLI: master

TonnyORG:
Expecting to fix #805, @kirrg001 please provide any feedback. I also thought in adding this line inside a `.then()` once the `systemctl stop ${this.systemdName}` had run, what do you think?

### DIFF
--- a/extensions/systemd/systemd.js
+++ b/extensions/systemd/systemd.js
@@ -35,6 +35,10 @@ class SystemdProcessManager extends ProcessManager {
             })
             .then(() => this.ui.sudo(`systemctl start ${this.systemdName}`))
             .then(() => this.ensureStarted({logSuggestion, socketAddress}))
+            .then(() => {
+                this.instance.config.set('bootstrap-socket', null);
+                return this.instance.config.save();
+            })
             .catch((error) => {
                 if (error instanceof CliError) {
                     throw error;
@@ -70,6 +74,10 @@ class SystemdProcessManager extends ProcessManager {
             })
             .then(() => this.ui.sudo(`systemctl restart ${this.systemdName}`))
             .then(() => this.ensureStarted({logSuggestion, socketAddress}))
+            .then(() => {
+                this.instance.config.set('bootstrap-socket', null);
+                return this.instance.config.save();
+            })
             .catch((error) => {
                 if (error instanceof CliError) {
                     throw error;


extensions/systemd/systemd.js
```diff
@@ -47,6 +47,8 @@ class SystemdProcessManager extends ProcessManager {
     stop() {
         this._precheck();
 
+        this.instance.config.set('bootstrap-socket', null);
```
vikaspotluri123: This won't persist changes, you need to run the `save` method as well 😄

TonnyORG: good catch, ty for your input. I feel like I should "delete" it once the promise for the `systemctrl stop` command has been resolved, what do you think?

```
        return this.ui.sudo(`systemctl stop ${this.systemdName}`)
            .then((response) => {
                this.instance.config.set('bootstrap-socket', null);
                return this.instance.config.save();
            })
            .catch((error) => {
                throw new ProcessError(error);
            });
```

vikaspotluri123: In terms of when the changes to the config should be reverted, I think it would be better to do this when the [socket](https://github.com/TryGhost/Ghost-CLI/blob/1d5dab17c23541eecfd5285c2da73a48fe4380e4/lib/utils/port-polling.js) is destroyed, because it's part of the cleanup and can help mitigate issues arising from `ghost stop` not being called. However, this might be a more relevant implementation for #787

acburdine: I agree with @vikaspotluri123 here - the original idea for #805 was to clean it up after ghost was started, which would mean adding this line in an extra `then` block [here](https://github.com/TryGhost/Ghost-CLI/blob/478277f641ff264ed23e53d3962ab9b14be5a1db/extensions/systemd/systemd.js#L37) and [here](https://github.com/TryGhost/Ghost-CLI/blob/478277f641ff264ed23e53d3962ab9b14be5a1db/extensions/systemd/systemd.js#L72).

Also, `config.set()` returns the config instance, so you can just use `config.set('bootstrap-socket', null).save();` 😄

TonnyORG: @acburdine ty for your input as well, I've refactorized the code to clean it up after ghost was started (start and restart commands). I kept the original `this.instance.config` call to make it consistent with the other portions where config is called in the file.

PR #849, opened 2018-10-21
Modify nginx template to redirect HTTP traffic to HTTPS

### LINKED ISSUE #826: Redirect HTTP -> HTTPS once SSL is setup
Ghost _should_ always be served over HTTPS. Particularly the admin panel. In the Ghost internals, we're working towards making HTTPS required in Ghost 3.0.

In the meantime, updating the CLI's behaviour will make for a smoother transition when the time comes.

If SSL is setup, the domain configured in Ghost should be HTTPS, and the nginx template for HTTP should be replaced with nginx configuration for a redirect:

```
server {
    listen 80;
    listen [::]:80;

    server_name <%= url %>;
    root <%= webroot %>;

    location ~ /.well-known {
        allow all;
    }

    return 301 https://<%= url %>$request_uri;
}
```

lunaticmonk:
Fixes #826 

@ErisDS Please review.

### DIFF
--- a/extensions/nginx/templates/nginx-ssl.conf
+++ b/extensions/nginx/templates/nginx-ssl.conf
@@ -23,4 +23,5 @@ server {
     }
 
     client_max_body_size 50m;
+    return 301 https://<%= url %>$request_uri;
 }


vikaspotluri123:
Hey @lunaticmonk 👋 thanks for your PR 😄 

The change would need to be made in the non-ssl template - The current added logic will redirect non-ghost URLs for sure, and possibly some Ghost URLs as well when they're already on HTTPS which could cause an infinite redirect loop

Another thing to consider: add logging to the CLI output / ask the user if they want to set up the SSL Redirect. This is new behavior which could create issues for some users

acburdine:
Also something to consider - this would only apply to new users of the CLI. Previous users would not get the same additions.

acburdine:
@lunaticmonk Going to close this PR due to inactivity - if you address the comments above then we can reopen this 😄

PR #850, opened 2018-10-30
feat(node): add support for node v10

acburdine:


kirrg001:
@allouis I've assigned you 👍

acburdine:
Node v10 tests were failing because ghost 2.4 wasn't released yet. Restarted the v10 tests, they should hopefully pass now.

PR #865, opened 2019-01-16
chore(start): show warning for insecure URLs in production

### LINKED ISSUE #863: Output a warning for production sites that don't use HTTPS
Not using HTTPS on your production site can leads to all manner of problems, such as posts not scheduling, navigation URLs misbehaving, the canonical URL being incorrect aaaand it's bad for SEO.

The whole world is moving away from HTTP, and we're the same. Our recommended stack is to use SSL, we _only_ serve SSL traffic on Ghost(Pro) and we'd like to do the same for self hosters.

To try to make this much, much clearer, `ghost start` and `ghost restart` should output a warning when Ghost is in production mode and the URL is not https.

> Warning: Ghost should always be configured with an HTTPS URL in production

The help link should be the (currently WIP) "How do I change my configured site URL?" FAQ

vikaspotluri123:
- closes #863 
- one test is failing because it's an edge case and I'm not sure if it should be handled

### DIFF
--- a/lib/commands/start.js
+++ b/lib/commands/start.js
@@ -34,8 +34,11 @@ class StartCommand extends Command {
 
         const url = instance.config.get('url');
 
-        if (this.system.environment === 'production' && url.indexOf('http://') >= 0) {
-            this.ui.log('Warning: Ghost should always be configured with an HTTPS URL in production\nFor more information, see our URL configuration FAQ at https://ghost.org/@TODO', 'yellow');
+        if (this.system.environment === 'production' && url.startsWith('http://')) {
+            this.ui.log([
+                'Using https on all URLs is highly recommended. In production, SSL is required when using Stripe.',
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
The failing test is

> does not show setup message after fresh LOCAL install without custom domain

i.e `http://12.23.34.55`

Should I add a check if the URL is an IP?

acburdine:
@vikaspotluri123 the messaging problably could be tweaked a bit to work for IP address http urls as well, e.g. “Ghost should always be configured with an HTTPS, non-IP url in production”

vikaspotluri123:
@ErisDS is this something that we want to do at this time, or should it be deferred to 5.0?

ErisDS:
This is definitely something I’d still like to do as we are going to add a similar warning into Ghost admin soon.

acburdine (approved):

PR #866, opened 2019-01-16

vikaspotluri123:
- update deps that aren't breaking and can be trusted
- regenerate lockfile to update subdeps (fixed unnecessary duplicate versions being required from the update)


acburdine:
thanks! will make sure to look at this later today

package.json
```diff
@@ -74,30 +74,30 @@
...
-    "eslint-plugin-ghost": "0.0.27",
+    "eslint": "5.12.0",
+    "eslint-plugin-ghost": "0.1.0",
     "has-ansi": "3.0.0",
     "mocha": "5.2.0",
     "nyc": "13.1.0",
     "proxyquire": "2.1.0",
     "sinon": "7.2.2",
-    "standard-version": "4.3.0",
+    "standard-version": "4.4.0",
```
acburdine: standard-version was kept at 4.3.0 for a reason - there was a change in 4.4.0 that keeps wanting to rewrite the entire change log.

Keeping it at 4.3 means we can pretty easily release without having to manually go and update the changelog.

vikaspotluri123: Rolling back!

PR #871, opened 2019-01-26
feat(yarn): use package-json instead of yarn for version resolution

acburdine:
While investigating/testing the update notification stuff for prerelease versions, I realized that the `package-json` package exists and that we already depend on it through the `latest-version` package. 

This PR removes some of the places where we were shelling out to the `yarn` command in favor of simply calling the `package-json` module. Simplifies the code and probably improves perf a bit due to not having to spin up yarn each time.

TODO:
- [x] fix tests


vikaspotluri123:
Ran some basic installation tests and they seem to work

PR #872, opened 2019-01-27
fix(install): ignore dotfiles except for .ghost-cli in dir check

### LINKED ISSUE #868: Ghost local install doesn't ignore dot files
### Issue Summary

When running `ghost local install` in a directory that contains a required dotfile; Installation fails with the message "Current directory is not empty, Ghost cannot be installed here." There are instance when these dot files are necessary or required by other systems. In my case when using Nodenv, Nodenv requires that a .node-version file be in the directory to map the correct local version of node that you would like to use.

### To Reproduce
Assuming that nodenv and the node version you would like to use is installed. Assuming that the ghost-cli is installed and working correctly. Assuming `CWD` is the selected install directory for your ghost site.

1. nodenv local [node-version-number]
2. ghost install local

> Message: Current directory is not empty, Ghost cannot be installed here.
> Debug Information:
>  OS: Mac OS X, v10.14.2
>  Node Version: v10.13.0
>  Ghost-CLI Version: 1.9.8
>  Environment: production
>  Command: 'ghost install local'
> Try running ghost doctor to check your system for known issues.
> You can always refer to https://docs.ghost.org/docs/troubleshooting#section-cli-errors for troubleshooting.

### Technical details:

* Ghost Version: 1.9.8
* Node Version: 10.13.0
* Browser/OS: n/a
* Database: n/a

### Workaround:
A simple work around is to make the node version global, but this defeats the purpose of having a version manager. Another workaround is to move the dot file one level higher; this causes an issue if you are deploying from a repo that use the files on a remote server.

It would be nice to have the ability via the command line to ignore dot files without removing them.

acburdine:
closes #868
- add dir-is-empty util that checks empty directories correctly

TODO:
- [x] fix install cmd tests

### DIFF
--- a/lib/utils/dir-is-empty.js
+++ b/lib/utils/dir-is-empty.js
@@ -1,0 +1,14 @@
+const fs = require('fs');
+
+const debugLogRegex = /^ghost-cli-debug-.*\.log$/i;
+const importantDotfiles = ['.ghost-cli'];
+
+module.exports = function dirIsEmpty(dir) {
+    const files = fs.readdirSync(dir);
+
+    if (!files.length) {
+        return true;
+    }
+
+    return files.every(file => file.match(debugLogRegex) || (file.startsWith('.') && !importantDotfiles.includes(file)));
+};
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -13,17 +13,15 @@ class InstallCommand extends Command {
     }
 
     run(argv) {
-        const fs = require('fs-extra');
-        const every = require('lodash/every');
         const errors = require('../errors');
         const yarnInstall = require('../tasks/yarn-install');
+        const dirIsEmpty = require('../utils/dir-is-empty');
         const ensureStructure = require('../tasks/ensure-structure');
 
         let version = argv.version;
-        const filesInDir = fs.readdirSync(process.cwd());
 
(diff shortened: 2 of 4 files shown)


lib/utils/dir-is-empty.js
```diff
@@ -0,0 +1,15 @@
...
+const importantDotfiles = ['.ghost-cli'];
+
+module.exports = function dirIsEmpty(dir) {
+    const files = fs.readdirSync(dir);
+
+    if (!files.length) {
+        return true;
+    }
+
+    return every(files, file => file.match(debugLogRegex) || (file.startsWith('.') && !importantDotfiles.includes(file)));
```
vikaspotluri123:
```suggestion
    return files.filter(file => file.match(debugLogRegex) || (file.startsWith('.') && !importantDotfiles.includes(file));
```

acburdine: realized actually that node 6 implements Array.every natively, meaning we don't even need the lodash dep here.

Also, `every` is a better choice for the logic here imo - it returns a boolean consistently and will exit early if one file is found that does not match the condition.

vikaspotluri123: Whoops, brain fart! I was thinking we're returning a file list 🤦‍♂️

acburdine: done 😄

PR #878, opened 2019-03-04
feat(pre-checks): add ability to skip pre-checks with env var

### LINKED ISSUE #867: Ensuring correct ~/.config folder ownership fails on shared hosting
### Summary

On a shared hosting we do not own root user. Ghost will fail during the install because it check that current user has right on the ~/.config folder.

I don't understand why because it don't seems to be required during the install.
I successfully install Ghost on this shared hosting by commenting this step.

### Steps to Reproduce

1. Having a user that doesn't not own the ~/.config folder.
2. Run `ghost install`

### Technical details

* OS: Debian
* Node Version: 8.10.0
* Ghost-CLI Version: 1.9.9
* Environment: production
* Command: `ghost install`

## Log

```bash
uid****@od-***:~/blog3$ ghost install
+ sudo chown -R ****_blog:****_blog /home/clients/*******/.config
? Sudo Password [input is hidden] 
```

I do not own the root password... Could we remove this check or making it optional?

acburdine:
closes #867
- add GHOST_CLI_PRE_CHECKS environment variable that can be
  set to 'false' to skip pre checks

### DIFF
--- a/lib/utils/pre-checks.js
+++ b/lib/utils/pre-checks.js
@@ -27,7 +27,8 @@ module.exports = function preChecks(ui, system) {
                     'yellow'
                 );
             }
-        })
+        }),
+        enabled: () => process.env.GHOST_CLI_PRE_CHECKS !== 'false'
     }, {
         title: 'Ensuring correct ~/.config folder ownership',
         task: () => fs.lstat(configstore).then((stats) => {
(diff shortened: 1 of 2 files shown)


acburdine:
note: https://github.com/TryGhost/Ghost-CLI/pull/879 should fix the breaking test

PR #889, opened 2019-04-16
Improve the CLI help output

ErisDS:
Whilst attempting to update our CLI docs, I found lots of places where the CLI output was very hard to read. We have a LOT of flags, so it can become really really tricky to understand what is / isn't possible when reading the help output.

Help should be the source of truth, so I've updated it to be a little clearer. None of these changes should be functional, except the expansion to having `--version1` as the long form of `--v1`.

Everything else should _only_ affect what help outputs.

The main thing is that global options are now shown in a "global options" group. The only options left in `Options:` are the command specific options. I've also tweaked the order, so that the commands are easier to read. 

---

A side note: in Gscan and other cli-based tools we use [@tryghost/pretty-cli](https://www.npmjs.com/package/@tryghost/pretty-cli), which wraps http://sywac.io/ instead of yargs. Sywac 1) is designed for dynamic command and option loading 2) has highly configurable, pretty, coloured help output 3) doesn't support the --no boolean switch and 4) doesn't seem to be actively maintained.

I think switching the CLI over to sywac would be a fairly involved task, It would solve some problems more neatly (dynamic loading, better output) and cause others (duplication of flags to reimplement --no switch) so prob not worth doing right now, just wanted to mention its existence!

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -142,8 +142,7 @@ const bootstrap = {
         }
 
         // Yargs magic
-        // TODO: evaluate if `wrap` is actually needed?
-        yargs.wrap(Math.min(100, yargs.terminalWidth()))
+        yargs.wrap(Math.min(150, yargs.terminalWidth()))
             .epilogue('For more information, see our docs at https://docs.ghost.org/api/ghost-cli/')
             .option('d', {
                 alias: 'dir',
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -144,33 +144,40 @@ const bootstrap = {
         // Yargs magic
         yargs.wrap(Math.min(150, yargs.terminalWidth()))
             .epilogue('For more information, see our docs at https://docs.ghost.org/api/ghost-cli/')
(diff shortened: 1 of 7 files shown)


ErisDS:
@acburdine I tested these changes as best as possible, they seem to work fine and the build passes locally. Can you cast a quick 👁 over the code changes to double check that nothing I've done will have wider impact than I realised, particularly the change in order of arguments to Object.assign() here: https://github.com/TryGhost/Ghost-CLI/pull/889/commits/25c740787114017267e432add50d4ab41acbf4c0

Fairly sure it should be fine!

acburdine:
looks like it's the `yarn audit` check that's failing the tests. May need to make a dependency update PR to fix that.

ErisDS:
Sure but we can do that later - right? The build failing isn't related to the changes.

acburdine:
fair point - I added it in so that I'd be able to see when dependencies were out of date. With renovator coming it's probably not as important.

ErisDS:
Updated commit messages to follow semantic commit.

PR #932, opened 2019-05-26
feat(nginx): customisable paths & program name

ErisDS:
no issue

- Respect environment variables that customise where nginx lives and what it is called
- Use env vars, not config, because this will be system-wide
- Allows CLI to work with "flavours" of nginx, like OpenResty

The two env vars are:

NGINX_PROGRAM_NAME
NGINX_CONFIG_PATH

I did wonder as part of this if `dpkg -l | grep xxx` is the best way to find out if something is installed, and whether using which, or a version check, might be more interoperable 🤔 (thinking about the case of symlinking openresty as nginx)

### DIFF
--- a/extensions/nginx/acme.js
+++ b/extensions/nginx/acme.js
@@ -7,6 +7,8 @@ const download = require('download');
 
 const {errors: {CliError, ProcessError, SystemError}} = require('../../lib');
 
+const nginxProgramName = process.env.NGINX_PROGRAM_NAME || 'nginx';
+
 function isInstalled() {
     return fs.existsSync('/etc/letsencrypt/acme.sh');
 }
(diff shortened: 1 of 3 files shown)


ErisDS:
Currently verifying this patch, seems to be working 👍

ErisDS:
Soooo this is definitely working as desired, however if you use OpenResty instead of nginx, there are a couple of gotchas, where openresty doesn't quite have the same setup as nginx out-of-the box.

Main thing is that instead of `/etc/nginx/` our config path will be `/usr/local/openresty/nginx/conf`. I tried a couple of different things to change this, but fighting it makes life harder & symlinks don't quite seem to work nicely either, so you'll want to set the following env vars:

```
export NGINX_PROGRAM_NAME=openresty
export NGINX_CONFIG_PATH=/usr/local/openresty/nginx/conf
```

(I put them in /etc/profile.d/ghostcli-openresty.sh, so they are global)

Once you've done that you'll want to create the following folders:

`sites-available`, `sites-enabled`, `snippets`, `conf.d` inside of `/usr/local/openresty/nginx/conf` owned by root and with perms 755. The rest of the standard nginx files are there, but these folders are missing.

Then you'll want to update `/usr/local/openresty/nginx/conf/nginx.conf`to have `include /usr/local/openresty/nginx/conf/sites-enabled/*;` instead of a server block...

If you want a default file for sites-enabled to test openresty is still working, this works:

```
server {
    listen       80;
    server_name  localhost;

    location / {
        root   html;
        index  index.html;
    }

    # redirect server error pages to the static page /50x.html
    #
    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
        root   html;
    }
}
```

You also need a `/var/www/` folder owned by root with perms 775.

Finally, once you're done with setting up, the last thing to do before running `ghost install` is always test that `sudo openresty -s reload` works cleanly, cos if there are issues, it borks the setup really late in the process!

PR #982, opened 2019-07-21
fix(config): allow root user to run config command

### LINKED ISSUE #981: Ghost CLI won't run in (community supported) ghost docker image
### Issue Summary

When run in Ghost community docker image, 'ghost' cli refuses to run (b/c ghost was installed as root).

### To Reproduce

1. Install Ghost in docker image from: https://hub.docker.com/_/ghost/
2. After some time, come back to make a configuration change in image:

Expect: Ghost CLI works

Observe:

```sh
$ docker exec -it docker_blog_1 sh
# ghost ls
You can't run commands as the 'root' user.
Switch to your regular user, or create a new user with regular account privileges and use this user to run 'ghost ls'.
For more information, see https://docs.ghost.org/install/ubuntu/#create-a-new-user-.
```

This behavior is not unexpected, it is documented.  But I'm raising it as a bug nevertheless because it's a *policy* not a *technical* restriction and (sez me) that it is inappropriate in a container environment.  

There, the container provides a sandbox for the rest of the host system, and the "root" user in the container has no more privilege than a regular user in the host system.  So even if Ghost is running as root in the container and the hacker exploits a code execution flaw, s/he can own all of Ghost but still cannot access the host system.

### Technical details:

* Ghost Version: 2.6.0
* Node Version: 10.15.3
* Browser/OS: Debian 9 (stretch)
* Database: sqlite3

## Bug submission checklist

Please fill out this checklist to acknowledge that you followed the requirements to submit a bug report.

- [x] Tried to find help in the forum & docs
- [x] Checked for existing issues
- [x] Attached log file
- [x] Provided technical details incl. operating system

acburdine:
closes #981
- this enables the config command to be run as the root user

### DIFF
--- a/lib/commands/config.js
+++ b/lib/commands/config.js
@@ -49,5 +49,6 @@ ConfigCommand.description = 'View or edit Ghost configuration';
 ConfigCommand.longDescription = '$0 config [key] [value]\n View or modify the configuration for a Ghost instance.';
 ConfigCommand.params = '[key] [value]';
 ConfigCommand.options = options;
+ConfigCommand.allowRoot = true;
 
 module.exports = ConfigCommand;


bobhy:
I like the intent of this change, but the default should probably be `false` to preserve backward compatibility.

acburdine:
@bobhy it is undefined on all commands by default, which given the way it's used is equivalent to false by default, see https://github.com/TryGhost/Ghost-CLI/blob/master/lib/command.js#L148

devinowen:
Hey team! Does this change allow you to run `ghost start` or `ghost run` as the root user? If so, how should I pass that param?

vikaspotluri123:
@devinowen this PR only allows the `config` command to be run as root, it doesn't modify the behavior of start / run

devinowen:
Thanks @vikaspotluri123!

Issue #460, opened 2017-08-22
Missing package(s): nginx and Local MySQL install not found


xolee:
hi, my host have installed nginx and mysql，how to config nginx and mysql directory?

```
ghost@VM-34-185-ubuntu:/home/siteList/blog.ghost$ ghost install
� Checking system Node.js version
� Checking current folder permissions
System checks failed with message: 'Missing package(s): nginx'
Some features of Ghost-CLI may not work without additional configuration.
For local installs we recommend using `ghost install local` instead.
? Continue anyway? Yes
� Checking operating system
Local MySQL install not found. You can ignore this if you are using a remote MySQL host.
Alternatively you could:
a) install MySQL locally
b) run `ghost install --db=sqlite3` to use sqlite
c) run `ghost install local` to get a development install using sqlite3.
? Continue anyway? No
� Checking MySQL is installed
A SystemError occurred.

Message: MySQL check failed.

ghost@VM-34-185-ubuntu:/home/siteList/blog.ghost$ ghost install --no-stack
✔ Checking system Node.js version
✔ Checking current folder permissions
ℹ Checking operating system [skipped]
Local MySQL install not found. You can ignore this if you are using a remote MySQL host.
Alternatively you could:
a) install MySQL locally
b) run `ghost install --db=sqlite3` to use sqlite
c) run `ghost install local` to get a development install using sqlite3.
? Continue anyway? No
✖ Checking MySQL is installed
A SystemError occurred.

Message: MySQL check failed.

Debug Information:
    Node Version: v6.11.2
    Ghost-CLI Version: 1.1.1
    Environment: production
    Command: 'ghost install --no-stack'

Please refer to https://docs.ghost.org/v1/docs/troubleshooting#section-cli-errors for troubleshooting.
```

acburdine:
@xolee Sorry for missing this!

Would you be able to run `which nginx` and `which mysqld` on your server? I want to double check that those commands return something.

klpek:
the same probelm.
I use ubuntu16.04 32bit.  Nodejs v6.11.3

ErisDS:
@klpek can you provide the information that @acburdine asked for above please? Can't fix this without knowing what your system expects 😊

klpek:
~$ nginx -V
nginx version: nginx/1.12.1
built by gcc 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4)
built with OpenSSL 1.0.2l  25 May 2017
[...]

PR #1029, opened 2019-10-25
Use `which` to detect packages rather than `dpkg`

heyakyra:
Currently, ghost assumes there is no MySQL or Nginx package installed
on non-Debian based GNU+Linux distributions. We fix this by querying
for packages using `which` instead of `dpkg`.

References #460

### DIFF
--- a/extensions/nginx/index.js
+++ b/extensions/nginx/index.js
@@ -281,7 +281,7 @@ class NginxExtension extends Extension {
 
     isSupported() {
         try {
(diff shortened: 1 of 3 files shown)


acburdine:
this will be superseded by #1313 - instead of using `which` or `dpkg` we're going to search the list of running services on the machine to determine if `nginx` is among them.

acburdine:
closing this since #1313 was merged

PR #1037, opened 2019-10-30
Added verbose yarn output

ErisDS:
@acburdine Can we add a tiny little fix like this and ship it to help with debugging https://github.com/TryGhost/Ghost/issues/11129?

I'm not sure if we also need to fix https://github.com/TryGhost/Ghost-CLI/blob/master/lib/utils/yarn.js to include stderr as mentioned in https://github.com/lovell/sharp/issues/1942#issuecomment-547513885 but that all seems like it needs a bit more time and attention. 

Definitely want to switch to outputting this info to a log file as well, but again, that takes time whereas this could help immediately.

### DIFF
--- a/lib/tasks/yarn-install.js
+++ b/lib/tasks/yarn-install.js
@@ -62,7 +62,11 @@ module.exports = function yarnInstall(ui, zipFile) {
     tasks.push({
         title: 'Installing dependencies',
         task: (ctx) => {
-            const observable = yarn(['install', '--no-emoji', '--no-progress'], {
+            const yarnOpts = ['install', '--no-emoji', '--no-progress'];
+            if (ui.verbose) {
+                yarnOpts.push('--verbose');
+            }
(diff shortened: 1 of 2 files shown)


lib/tasks/yarn-install.js
```diff
@@ -62,7 +62,11 @@ module.exports = function yarnInstall(ui, zipFile) {
     tasks.push({
         title: 'Installing dependencies',
         task: (ctx) => {
-            const observable = yarn(['install', '--no-emoji', '--no-progress'], {
+            const yarnOpts = ['install', '--no-emoji', '--no-progress'];
+            if (ui.verbose) {
+                yarnOpts.push('verbose');
```
acburdine: wouldn't this need to be:
```suggestion
                yarnOpts.push('--verbose');
```
?

ErisDS: Ha yes, updated

PR #1042, opened 2019-11-04
feat(proxy): add support for cli behind proxy

cheahkhing:
add proxy-agent to get proxy settings
make sure all requests are handled properly with proxy agent
fix all unit tests failures


lib/utils/get-proxy-agent.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
acburdine: we don't need to add `use strict` to the top of files anymore

lib/utils/get-proxy-agent.js
```diff
@@ -0,0 +1,17 @@
...
+
+const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
```
acburdine: can this use destructuring? 
```suggestion
const {getProxyForUrl} = require('proxy-from-env');
```

lib/utils/get-proxy-agent.js
```diff
@@ -0,0 +1,17 @@
...
+var HttpsProxyAgent = require('https-proxy-agent');
```
acburdine: this should use const:
```suggestion
const HttpsProxyAgent = require('https-proxy-agent');
```

vikaspotluri123: I think we should also lazy require this since it won't be used most of the time?

acburdine: 👍 good point, definitely agree

lib/utils/get-proxy-agent.js
```diff
@@ -0,0 +1,15 @@
+const {getProxyForUrl} = require('proxy-from-env');
+const HttpsProxyAgent = require('https-proxy-agent');
+const NPM_REGISTRY = 'https://registry.npmjs.org/';
+
+let proxyAgent = false;
+
+module.exports = function getProxyAgent() {
+    // Initialize Proxy Agent for proxy support if needed
+    const proxyAddress = getProxyForUrl(NPM_REGISTRY);
+    if (proxyAddress) {
```
acburdine: like @vikaspotluri123 said, can we move the require for the `https-proxy-agent` dependency down inside this function? that way, the agent is only ever required if it's used, and since we cache the proxy agent itself in theory the require should only happen once.

PR #1130, opened 2020-03-18
Switched testing to GitHub Actions

daniellockyer:
no issue


test/unit/ui/index-spec.js
```diff
@@ -14,6 +14,11 @@ const execa = require('execa');
 const modulePath = '../../../lib/ui';
 
 describe('Unit: UI', function () {
+    before(() => {
+        // Workaround because GitHub Actions doesn't currently report a TTY
+        process.stdout.isTTY = true;
```
acburdine: this is probably as good of a workaround as any, seems fine to me

PR #1164, opened 2020-04-22
Fixed name

therecluse26:
The name Ghost-CLI is a huge missed opportunity, so I decided to fix it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,9 @@
 
 ## Project Goals
 
-The objective of the Ghost CLI project is to make setting up and maintaining a Ghost site as straight forward as possible for people who do not want to use Ghost(Pro).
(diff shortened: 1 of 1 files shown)


acburdine:
@therecluse26 thank you for the suggestion - we actually titled a release of Ghost-CLI similarly in the past :smile: (https://github.com/TryGhost/Ghost-CLI/releases/tag/1.0.0-alpha.8)

That said, we won't be changing the name of the project - it was chosen deliberately.

brad-tilmor:
@acburdine I am both filled with the fury of 10,000 suns and totally understanding of your decision

therecluse26:
@acburdine I am both filled with the fury of 10,000 suns and totally understanding of your decision

ErisDS:
To be fair though, where it doesn't hamper "typability" or "readability" we do love a good easter egg. The port number was chosen with great care, and there's fun to be found in `ghost --help` among other places 😬

PR #1171, opened 2020-04-28
Fix for --no-prompt sudo shell execution in Dockerfile of ghost

iamkhalidbashir:
This is a fix if u run 'ghost setup nginx' in Dockerfile of ghost. 

**Problem**
- when 'RUN gosu node ghost setup nginx' is added in Dockerfile to set ghost on docker, following error comes out:-
```
/usr/local/lib/node_modules/ghost-cli/lib/ui/index.js:148
            throw new errors.SystemError('Prompts have been disabled, all options must be provided via command line flags');
```

- Reason for this is because **this.ui.sudo(command)** called in https://github.com/TryGhost/Ghost-CLI/blob/019f4da8ac699cf68d8cc7ef03c836df982cd501/extensions/nginx/index.js#L112

- sudo(command) in **ui** then tries to call **this.prompt()** which results in an error because docker doesn't allow prompt to open

**Solution**
The user can set an environment variable of SUDO_PASSWORD and then this will automatically enter the password in the sudo prompt.

### DIFF
--- a/lib/ui/index.js
+++ b/lib/ui/index.js
@@ -246,7 +246,7 @@ class UI {
             /* istanbul ignore else */
             // If sudo has prompted for a password, then it will be the last line output
             if (lines[lines.length - 1] === prompt) {
-                if(process.env.SUDO_PASSWORD !== undefined){
+                if (process.env.SUDO_PASSWORD !== undefined) {
                     this.prompt({
                         type: 'password',
                         name: 'password',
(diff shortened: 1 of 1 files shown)


acburdine:
@iamkhalidbashir thank you for the submission! unfortunately though, I don't think this is behavior we want to add 😕 

I mentioned it in [this comment](https://github.com/docker-library/ghost/issues/156#issuecomment-428159861) on the docker-library/ghost repo, but essentially, Ghost-CLI isn't intended or expected to work fully within a Docker environment.

Having said that, there's a way to achieve what you're looking to do _without_ adding these changes, using passwordless sudo. Something like [this](https://stackoverflow.com/a/48329093/3797605) might do the trick.

Issue #1201, opened 2020-05-22
Improve IPC messaging to make errors more visible


ErisDS:
**Problem:**

Frequently, Ghost fails to start with nothing more than an error code to show for itself, and the error code is often just "0". This makes debugging really difficult, people new to the tool don't have a chance, and even when they do make it to the forum it's really hard for seasoned Ghost-CLI users to determine where to go looking.

Here's a handful of examples from issues and the forum:
- https://github.com/TryGhost/Ghost-CLI/issues/1103
- https://github.com/TryGhost/Ghost-CLI/issues/1120
- https://github.com/TryGhost/Ghost-CLI/issues/1070
- https://forum.ghost.org/t/ghost-installation-rebuilding-issues/11443
- https://forum.ghost.org/t/latest-ghost-wont-start-openbsd-6-7/

This seems to happen mostly on local installs, but I'm not convinced that's entirely true. Other commands can sometimes surface a partial error:

- https://github.com/TryGhost/Ghost-CLI/issues/1150
- https://forum.ghost.org/t/attempt-to-upgrade-2-to-3-fails/13170

But still not enough to debug an issue.

**Background:**

Whenever anyone runs `ghost start` either manually, or as the last step in `ghost install`, 3 separate processes are created:

- `ghost start`
- `ghost run` 
- `node index` (Ghost itself)

Often, the error happens in Ghost or the `ghost run` command, and isn't passed up. Therefore `ghost run -D` can sometimes show the error, but sometimes you have to go digging in the Ghost logs!

**Solution:**

We need to make sure that any and all relevant error messages are passed to the top level process and displayed, no matter which process they occurred in.

Ideally we also need to output which process the error occurred in, and if possible the debug info from _that_ process. 

Ghost was recently updated so that it passes additional debug info: https://github.com/TryGhost/Ghost/blob/21d38b2d37d9dd1797a40f5d14bfa600b183581f/core/server/ghost-server.js#L346

The `ghost run` process may need a similar update to provide some more debug info. 

This is a somewhat nebulous and hard-to-reproduce problem, but we see it all the time. I think the first step is to assess how the IPC is working, make sure errors are passed around, and then iterate on improving the output.

vikaspotluri123:
The change that was published over the weekend seems to be on the right path. If people are still running into the `exit code 0` issue, we'll hopefully have more debug information to improve the CLI.

More references:

https://forum.ghost.org/t/ghost-error-occurred-knex-sqlite/14669

[...]

PR #1202, opened 2020-05-22
fix(local-process) output Ghost error message when process exits

vikaspotluri123:
Tests are broken, but I'd like to get a logic review first 😄

### DIFF
--- a/lib/utils/local-process.js
+++ b/lib/utils/local-process.js
@@ -34,15 +34,21 @@ Ensure the content folder has correct permissions and try again.`));
         }
 
         return new Promise((resolve, reject) => {
+            let errorBuffer = '';
+
             const cp = childProcess.spawn('node', [process.argv[1] , 'run'], {
                 cwd: cwd,
                 detached: true,
-                stdio: ['ignore', 'ignore', 'ignore', 'ipc'],
+                stdio: ['ignore', 'ignore', 'pipe', 'ipc'],
                 env: Object.assign({}, process.env, {NODE_ENV: environment})
             });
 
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
One concern I have with this implementation is I'm not sure how the pipe will work after disconnecting the CP - will it still be connected on the CLI side (I would presume no), and where will stderr go on the cp-side?

lib/utils/local-process.js
```diff
@@ -65,6 +74,9 @@ Ensure the content folder has correct permissions and try again.`));
 
                 /* istanbul ignore else */
                 if (msg.started) {
+                    buffer = '';
```
acburdine, line 77: do we need to reset buffer here? seems like it shouldn't matter 🤔

acburdine:
> One concern I have with this implementation is I'm not sure how the pipe will work after disconnecting the CP - will it still be connected on the CLI side (I would presume no), and where will stderr go on the cp-side?

it will likely still be connected, we might need to call `cp.stderr.destroy()` to close the stream. On the cp side, I _believe_ the stderr will still be ignored? According to the node [stream docs](https://nodejs.org/api/stream.html#stream_readable_destroy_error):
> After this call, the readable stream will release any internal resources and subsequent calls to push() will be ignored.

acburdine:
@vikaspotluri123 created a small repro case to answer your above questions: https://gist.github.com/acburdine/f6cf5241281ee81bfee1bab4214363de

vikaspotluri123:
Forgot to add the issue ref - refs #1201

PR #1207, opened 2020-05-28
Added systemd migration to store node version in instance config

vikaspotluri123:
Wanted to get an initial review if this is the right track. I'll move stuff, lint, test, etc. afterwards 😄

### DIFF
--- a/extensions/systemd/index.js
+++ b/extensions/systemd/index.js
@@ -1,4 +1,4 @@
-'use strict';
+// @ts-check
 
 const fs = require('fs-extra');
 const path = require('path');
@@ -10,6 +10,17 @@ const {Extension, errors} = require('../../lib');
 const {ProcessError, SystemError} = errors;
 
 class SystemdExtension extends Extension {
+    migrations() {
+        const migrations = require('./migrations');
+
+        return [{
+            before: '1.15.0',
+            title: 'Saving node binary',
+            skip: ({instance}) => instance.process.name !== 'systemd' || instance.nodeBinaryIsKnown(),
+            task: migrations.saveNodeExecPath
+        }];
+    }
+
     setup() {
         return [{
             id: 'systemd',
--- a/lib/instance.js
+++ b/lib/instance.js
@@ -124,7 +124,7 @@ class Instance {
      * @public
      */
     get nodeBinary() {
-        return this._cliConfig.get('node-binary', process.argv[0]);
+        return this._cliConfig.get('node-binary', process.execPath);
     }
     set nodeBinary(value) {
         this._cliConfig.set('node-binary', value).save();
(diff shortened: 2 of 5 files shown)


acburdine:
@vikaspotluri123 approach makes sense here 👍

extensions/systemd/index.js
```diff
@@ -10,6 +10,26 @@ const {Extension, errors} = require('../../lib');
...
+            before: '1.15.0',
+            title: 'Saving node binary',
+            skip: ({instance}) => instance.process.name !== 'systemd' || instance.nodeBinaryIsKnown(),
+            task: async ({instance}) => {
+                const serviceFilename = `ghost_${instance.name}.service`;
+                const fullFilePath = path.resolve(instance.dir, './system/files', serviceFilename);
+
+                const fileContents = await fs.readFile(fullFilePath, 'utf8');
+
+                const nodeBinaryLine = fileContents.split('\n').filter(line => line.startsWith('ExecStart'))[0];
```
acburdine, line 25: minor efficiency thing
```suggestion
                const nodeBinaryLine = fileContents.split('\n').find(line => line.startsWith('ExecStart'));
```

extensions/systemd/index.js
```diff
@@ -10,6 +10,26 @@ const {Extension, errors} = require('../../lib');
...
+                const nodeBinary = nodeBinaryLine.slice(nodeBinaryLine.indexOf('=') + 1, nodeBinaryLine.indexOf(' '));
+
+                instance.nodeBinary = nodeBinary;
```
acburdine, line 28:
```suggestion
                const [, nodeBinary] = nodeBinaryLine.split('=');

                instance.nodeBinary = nodeBinary.trim();
```

vikaspotluri123: The line looks like this:

`ExecStart=/usr/bin/node /usr/bin/ghost run`

Would you prefer `nodeBinary.split(' ').shift()`?

acburdine: oh whoops - forgot it was put like that...

yeah in that case `nodeBinary.trim().split(' ').shift()` should work (i'd keep the `trim` just in case there's a space between the `=` and the `/usr/bin/node`

vikaspotluri123:
@acburdine I think this is good to review / merge individually. I'm going to work on the binary check changes in a separate branch.

lib/instance.js
```diff
@@ -116,6 +116,30 @@ class Instance {
...
 
+    /**
+     * Node Binary path used when installing this version
+     *
+     * @property nodeBinary
+     * @type string
+     * @public
+     */
+    get nodeBinary() {
+        return this._cliConfig.get('node-binary', process.argv[0]);
```
acburdine, line 127: would it make sense to use process.execPath instead?

vikaspotluri123: Yep, it looks like `process.execPath === process.argv[0]`

PR #1210, opened 2020-06-03
feat(instance): add method for retrieving all available configs

acburdine:


lib/instance.js
```diff
@@ -245,6 +246,29 @@ class Instance {
         this.system.setEnvironment(env === 'development', setNodeEnv);
     }
 
+    /**
+     * Gets a list of availble configurations for this instance
+     *
+     * @method getAvailableConfigs
+     * @return {Promise<object>}
```
vikaspotluri123, line 253:
```suggestion
     * @return {Promise<{[s: string]: Config}>}
```

I _think_ this should work as valid jsdoc

acburdine: yup - thanks for the suggestion 😄

PR #1216, opened 2020-06-05
Attempt to load binary modules when running binary checks

### LINKED ISSUE #1205: Improve Binary Dependency Check
Upgrading node is a common usecase, as a fact of how this works, it blows away binary dependencies and they have to be re-installed.

We have a binary dependency check in place that tries to detect if node changed, but this isn't catching all cases.

We need to catch the case that the binaries are out of date some how, and show the same error:

ref: https://forum.ghost.org/t/ghost-error-occurred-knex-sqlite/14669

vikaspotluri123:
Depends on #1207, closes #1205

Wanted to get a functional review before working on the tests 😊. The copy needs work, but I figured we can worry about that once the architecture is stable

### DIFF
--- a/lib/commands/doctor/checks/binary-deps.js
+++ b/lib/commands/doctor/checks/binary-deps.js
@@ -1,20 +1,91 @@
-const chalk = require('chalk');
-const {SystemError} = require('../../../errors');
+// @ts-check
+
+const path = require('path');
+const {SystemError, ProcessError} = require('../../../errors');
+const {shouldUseGhostUser} = require('../../../utils/use-ghost-user');
 
 const taskTitle = 'Checking binary dependencies';
 
-function binaryDeps(ctx, task) {
+async function binaryDeps(ctx, task) {
     if (!ctx.instance) {
         return task.skip('Instance not set');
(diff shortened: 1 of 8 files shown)


lib/commands/doctor/checks/binary-deps.js
```diff
@@ -1,22 +1,93 @@
-const chalk = require('chalk');
-const {SystemError} = require('../../../errors');
+// @ts-check
```
vikaspotluri123, line 1: This is the main file that needs to be reviewed, most everything else is from #1207

precheck.js
```diff
@@ -0,0 +1,23 @@
...
+};
+
+for (const binaryDependency of dependencies) {
+    result[binaryDependency] = {};
+    try {
+       require(binaryDependency);
+       result[binaryDependency].success = true;
+    } catch (error) {
+       result[binaryDependency].success = false;
+       result[binaryDependency].error = error.toString();
```
vikaspotluri123, line 19: If I don't do `error.toString()` here, most of the useful information is stripped

PR #1217, opened 2020-06-07
[SSL] Add -dsaparam to speed up the generation of encryption keys

jilljenn:
Hi, the generation key was taking too long for me, so I checked on the Internet and found [this StackOverflow answer](https://security.stackexchange.com/a/95184/8965):

> This is considerably faster because it does not need to nest the primality tests, and thus only thousands, not millions, of candidates will be generated and tested.  
> As far as academics know, DSA-like parameters for DH are equally secure; there is no actual advantage to using "strong primes" (the terminology is traditional and does not actually imply some extra strength).

### DIFF
--- a/extensions/nginx/index.js
+++ b/extensions/nginx/index.js
@@ -185,7 +185,7 @@ class NginxExtension extends Extension {
         }, {
             title: 'Generating Encryption Key (may take a few minutes)',
             skip: () => fs.existsSync(dhparamFile),
-            task: () => this.ui.sudo(`openssl dhparam -out ${dhparamFile} 2048`)
+            task: () => this.ui.sudo(`openssl dhparam -dsaparam -out ${dhparamFile} 2048`)
                 .catch(error => Promise.reject(new ProcessError(error)))
         }, {
             title: 'Generating SSL security headers',


acburdine:
@jilljenn hey - apologies for not looking at this yet.

I did a bit more research (including reading through that StackOverflow post) and unfortunately I'm not familiar enough with encryption configurations to truly understand the implications of this change. In the stackoverflow answer you linked, it has a link to the [openssl docs](https://www.openssl.org/docs/man1.1.1/man1/dhparam.html#OPTIONS), which say:

> Beware that with such DSA-style DH parameters, a fresh DH key should be created for each use to avoid small-subgroup attacks that may be possible otherwise.

Currently, we only generate one dhparam file for the entire system, instead of one for each site installed on the system. This wording makes me thing that if we switch to use this `-dsaparam` option, we should change the nginx setup to generate one per site.

The _other_ interesting thing I found is that Mozilla's ssl-configs standards site (https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.4) seems to have a static dhparam key, so maybe it's not as big of a deal to add the `-dsaparam` flag 🤔

Ultimately, I need to do a bit more research before allowing this change.

acburdine:
closing in favor of #1354, which is essentially the same thing. ultimately based on the various research I did this change shouldn't affect things too much. Apologies for the delay.

Issue #1116, opened 2020-02-29
Theme demo post prompt has version number hardcoded


llakssz:
https://github.com/TryGhost/Ghost-CLI/blob/c3c7a55316df90fac0a5292291df67399df0377e/lib/tasks/major-update/ui.js#L50
https://github.com/TryGhost/Ghost-CLI/blob/c3c7a55316df90fac0a5292291df67399df0377e/lib/tasks/major-update/ui.js#L90

I guess just change it to:
'Visit the demo post at ${chalk.cyan(demoLink)} to see how your theme looks like.' ~~in Ghost 2.0~~

acburdine:
good catch - I'll make sure to get that updated.

PR #1226, opened 2020-06-08
fix(major-update): make theme demo prompt dynamic

acburdine:
closes https://github.com/TryGhost/Ghost-CLI/issues/1116


acburdine:
that's a good point - probably need to create an issue to revisit the major-update flow altogether 🤔

Issue #1218, opened 2020-06-07
Add option to force ghost install


bermudalocket:
### Summary

As a security precaution, `ghost` cannot be installed in a non-empty directory. This behavior is detrimental in at least one case. Consider a system utilizing `nodeenv` to manage multiple versions of NodeJS. If the system's global NodeJS version is set to 14.x, `ghost` cannot be installed due to version incompatibility. Creating a new directory for `ghost` and setting the local (directory) NodeJS version to 12.x via `nodeenv` creates a folder named `env`, which then blocks `ghost` from installing due to a non-empty directory.

Proposed solution is to add a `--force` flag to `ghost install`.

acburdine:
@bermudalocket that approach makes sense - I'll work on adding that 👍

ErisDS:
Sorry to throw a spanner in the works here, but whilst I agree that this report constitutes a genuine problem, I completely disagree that adding a `--force` flag is the correct solution.

1) because I had thought we already had code to ignore `.git` folders & that this could be added to that although I cannot find it now 🤔 

2) because we already have a `--force` flag in Ghost-CLI with a clear meaning.  `--force` is used to permit a re-run of a command that has already been run. Therefore, I would expect `ghost install --force` to specifically rerun a complete install in a directory where Ghost is already installed.

3) because we already have a flag pattern for skipping certain checks, e.g. `--no-check-mem` and so I would expect a flag that explicitly skipped a single check to follow this pattern, e.g. `--no-check-empty`.

When adding flags or commands, we have to consider how their name fits into the wider tool, in order to create a consistent toolset that's easy to work with and predict what will happen.

vikaspotluri123:
> 1. because I had thought we already had code to ignore `.git` folders & that this could be added to that although I cannot find it now 🤔

Here's the line you're looking for:
https://github.com/TryGhost/Ghost-CLI/blob/4f7c80c885351dc59897b6297aac0b0e43690e4a/lib/utils/dir-is-empty.js#L13

I guess the issue is nodeenv creates a non-dot folder which is why the CLI was complaining. I totally see what you're saying about 2 and 3 😄

acburdine:
Those points make sense, I’ll rename the flag 👍🏻

bermudalocket:
As an end user unfamiliar with the inner workings of this project the semantics of the flag makes no difference to me. If there's a choice that better adapts to an existing naming scheme then that is obviously the better choice.

straube:
[...]

PR #1237, opened 2020-06-28
fix(install): rename force flag to 'check-empty'

acburdine:
closes https://github.com/TryGhost/Ghost-CLI/issues/1218

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -23,7 +23,7 @@ class InstallCommand extends Command {
         let version = argv.version ? `${argv.version}` : null;
 
         // Check if the directory is empty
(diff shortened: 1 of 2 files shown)


lib/commands/install.js
```diff
@@ -23,7 +23,7 @@ class InstallCommand extends Command {
         let version = argv.version ? `${argv.version}` : null;
 
         // Check if the directory is empty
-        if (!dirIsEmpty(process.cwd()) && !argv.force) {
+        if (!dirIsEmpty(process.cwd()) && argv['check-empty']) {
```
vikaspotluri123, line 26: We could have a minor perf improvement by swapping these checks

PR #1260, opened 2020-07-29
Add dbport option for db setup

### LINKED ISSUE #1135: Add an option to specify MySQL DB Port
When installing Ghost there isn't an option to specify a custom database port for MySQL. If you're installing it on a managed database service that runs on a non-standard MySQL port this means you have to install ghost initially and let it fail, then manually edit the `config.production.json` file and add the port, then restart ghost.

Charizard:
Fixes #1135


lib/tasks/configure/get-prompts.js
```diff
@@ -54,6 +54,16 @@ module.exports = function getPrompts(config, argv, environment) {
             });
         }
 
+        if (!argv.dbport) {
```
acburdine, line 57: I don't think we need this option here - the database port is something that you generally don't have to configure for most Ghost-CLI setups, and as such adding a prompt for it only adds unnecessary steps that the user has to follow in order to setup a blog successfully. Just adding the dbport option in the file below this should be enough for this PR.

acburdine:
@Charizard I went ahead and remove the extra prompt so this could be mergeable, thanks for the contribution!

acburdine (approved):

Charizard:
@acburdine Thanks 🙏  Sorry, was not able to get to this.

PR #1284, opened 2020-08-24
feat(doctor): add free disk space check

acburdine:


lib/commands/doctor/checks/free-space.js
```diff
@@ -0,0 +1,34 @@
...
+
+    const disks = await sysinfo.fsSize();
+
+    // filter out disks with matching mount points, then sort in descending order by mount point length
+    // to get the mount point with greatest specificity
+    const [disk] = disks.filter(d => dir.startsWith(d.mount)).sort((a, b) => b.mount.length - a.mount.length);
+
+    if (!disk) {
+        // couldn't find a matching disk, early return
+        // TODO: maybe throw a warning of some sort here?
```
vikaspotluri123, line 20: We could have a warning that says `Unable to determine available free space. We suggest you have at least 1gb of space available.` with a 2s delay 🤔

lib/commands/doctor/checks/free-space.js
```diff
@@ -0,0 +1,34 @@
+const sysinfo = require('systeminformation');
+const {SystemError} = require('../../../errors');
+
+const MB_IN_BYTES = 1048576;
+
+// one version of Ghost takes ~400mb of space currently, to be safe we're going to say 1gb
+const MIN_FREE_SPACE = 1024;
+
+async function checkFreeSpace(ctx) {
+    const dir = ctx.instance ? ctx.instance.dir : process.cwd();
```
vikaspotluri123, line 10: Can't wait to `ctx.instance?.dir ?? process.cwd()` in the future

lib/commands/doctor/checks/free-space.js
acburdine: eventually I want to have the doctor command actually distinguish between "warnings" and "errors" - left this comment to remember to add a warning here whenever that distinction is possible.

lib/commands/doctor/checks/free-space.js
acburdine: 💯

PR #1309, opened 2020-10-06
Node 14 Support (Experimental)

acburdine:


test/unit/tasks/yarn-install-spec.js
```diff
@@ -12,6 +12,10 @@ const modulePath = '../../../lib/tasks/yarn-install';
 const errors = require('../../../lib/errors');
 
 describe('Unit: Tasks > yarn-install', function () {
+    beforeEach(() => {
+        process.env = {};
```
vikaspotluri123, line 16: Do we have any tests that depend on `NODE_ENV=test` (or something like that)

acburdine: Not that I'm aware of - the only NODE_ENV usage I recall seeing was when we set `NODE_ENV = 'production'` when calling `yarn install `, and in that case it's a hardcoded value.

lib/utils/local-process.js
```diff
@@ -44,7 +44,7 @@ Ensure the content folder has correct permissions and try again.`));
             });
 
             // Stick the pid into the pidfile so we can stop the process later
-            fs.writeFileSync(path.join(cwd, PID_FILE), cp.pid);
+            fs.writeFileSync(path.join(cwd, PID_FILE), `${cp.pid}`);
```
vikaspotluri123, line 47: Do you know what broke here?

vikaspotluri123 (approved):
LGTM 👍🏾

lib/utils/local-process.js
acburdine: Yeah so from what I was seeing Node 14 tightened up the type checking for the writeFile methods - it now errors if you pass it contents that aren't either a string or some sort of data buffer

vikaspotluri123: Aah I see. Could we do this to make it more obvious?

```suggestion
            fs.writeFileSync(path.join(cwd, PID_FILE), cp.pid.toString());
```

acburdine: Yeah the `toString` call makes sense - I'll make that change in the morning before I merge 🙂

PR #1310, opened 2020-10-07
Update E2E tests to use test matrix

vikaspotluri123:
Based on #1307, just testing some stuff for now :smile:


vikaspotluri123:
@acburdine can you take a look at the [mysql8 error](https://github.com/TryGhost/Ghost-CLI/runs/1218060905?check_suite_focus=true)? Not sure if there's something I'm doing wrong or if there's an actual bug 😅

acburdine:
So the error you're seeing is actually the majority of the reason for the MySQL 8 PR in the first place 😅. MySQL 8 defaults to using `caching_sha2_password` as the default password plugin, and the node MySQL driver doesn't yet support that plugin.

My guess is the GitHub action you're using is passing those inputs to the underlying docker image, and b/c the root user is created using the `caching_sha2_password` plugin, Ghost fails to connect to it 😬

Also the local install skips the MySQL setup step entirely, so this test I don't _think_ will really test any of the changes I added unfortunately.

vikaspotluri123:
I guess that's blocked by https://github.com/mirromutth/mysql-action/pull/7 then 🤔 assuming Github supports it, is there any issue with just manually creating / destroying the container?

acburdine:
Not that I can think of, unless it becomes a lot of work to maintain 🤔

vikaspotluri123:
> Also the local install skips the MySQL setup step entirely

Should we do this if the database is MySQL?

vikaspotluri123:
@acburdine I'm at a loss for what should be done for the error that's coming from MySQL. As previously, can't tell if it's a config issue or bug

https://github.com/TryGhost/Ghost-CLI/runs/1221918494?check_suite_focus=true#step:9:78

Aside: The `--dbport` flag is passed but not currently used by the CLI. There's a PR to support it

vikaspotluri123:
Fun fact: it looks like Ubuntu 20 comes w/ MySQL 8. Makes sense but I didn't realize. I'll leave some comments explaining some cool stuff I found, but this is _officially_ ready for review!

.github/workflows/e2e-test.yml
```diff
@@ -6,85 +6,38 @@ on:
...
 
 jobs:
-  ubuntu-16:
-    runs-on: ubuntu-16.04
-    name: 'Ubuntu 16.04'
+  e2e:
+    runs-on: ${{ matrix.os }}
+    name: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
```
vikaspotluri123, line 13: I think this is true by default, but I set it anyway

.github/workflows/e2e-test.yml
```diff
@@ -6,85 +6,38 @@ on:
...
+      matrix:
+        os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
+        password: [root]
+        node: [12.x]
```
vikaspotluri123, line 17: The cool thing about matricies is we can use some keys as variables - in this case, when updating node, we just have to look for the node version here instead of in the setup node action

.github/workflows/e2e-test.yml
```diff
@@ -6,85 +6,38 @@ on:
...
-      - uses: actions/setup-node@v1
-        with:
-          node-version: '12.x'
-      - run: cd cli && yarn && sudo yarn link --link-folder /usr/bin/
-      - name: Setting up Ghost instance
+          node-version: ${{ matrix.node }}
+      - name: Start MySQL service
+        run: sudo systemctl start mysql.service
+      - name: Fix MySQL Password
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
```
vikaspotluri123, line 32: Github has a web IDE for editing workflow files, boasting a decent level of maturity. It has support for auto complete of keys in many places which is really cool. I got burned a few times because of variable interpolation errors, though.

acburdine: So one comment here is that `if` keys don't need variable interpolation, so you _should_ be able to do 
```suggestion
        if: matrix.os == "ubuntu-20.04"
```

acburdine (approved):

Issue #308, opened 2017-07-07
You are able to skip systemd, but you are actually not


kirrg001:
If you run `ghost install` and you skip systemd setup, you won't be able to start Ghost, as linux & systemd setup is required, because you will run into

> Message: Systemd process manager has not been set up. Run `ghost setup linux-user systemd` and try again.

[you have to use `--process local` explicit if want to use the local process manager]. 

Should we remove the prompt for systemd? Or change the prompt to e.g. "Are you ready to setup systemd?". Or any other ideas?

I am still confused about this step. Both @AileenCGN and @cobbspur find that confusing as well.
It was also mentioned, that people might not even know what systemd means.

vikaspotluri123:
This should be fixed w/ #1311 - process manager is reverted to `local` if user says no to setting up systemd

PR #1311, opened 2020-10-07
feat(extensions): add `user skipped` task hook

vikaspotluri123:
refs #308, #1115

I'm not sure if this is the best way to do this, but it's the most pluggable thing that came to mind. I haven't fully tested this or ensured that changing the process manager the way I did will actually work 😄 It's best to look at 28a85b8 individually since it's the only logical change.

@acburdine can you please take a cursory look at this?

### DIFF
--- a/extensions/systemd/index.js
+++ b/extensions/systemd/index.js
@@ -14,7 +14,8 @@ class SystemdExtension extends Extension {
         return [{
             id: 'systemd',
             name: 'Systemd',
-            enabled: ({instance, argv}) => !argv.local && instance.config.get('process') === 'systemd',
+            enabled: ({instance, argv}) => !argv.local &&
+                 (instance.config.get('process') === 'systemd' || (argv.stages && argv.stages.includes('systemd'))),
             task: (...args) => this._setup(...args),
             skip: ({instance}) => {
                 if (fs.existsSync(`/lib/systemd/system/ghost_${instance.name}.service`)) {
@@ -22,16 +23,26 @@ class SystemdExtension extends Extension {
                 }
 
                 return false;
+            },
+            onUserSkip: ({instance, ui}) => {
+                ui.log('Systemd setup skipped, reverting to local process manager', 'yellow');
+                instance.config.set('process', 'local').save();
             }
         }];
     }
 
-    _setup({instance}, task) {
+    _setup({instance, ui}, task) {
         const uid = getUid(instance.dir);
 
         // getUid returns either the uid or null
         if (!uid) {
             return task.skip('The "ghost" user has not been created, try running `ghost setup linux-user` first');
+        }
+
+        if (instance.config.get('process') !== 'systemd') {
+            const currentProcessManager = instance.config.get('process');
+            ui.log(`Changing process manager from ${currentProcessManager} to systemd`, 'yellow');
+            instance.config.set('process', 'systemd').save();
         }
 
         const serviceFilename = `ghost_${instance.name}.service`;
--- a/lib/commands/setup.js
+++ b/lib/commands/setup.js
@@ -141,6 +141,10 @@ class SetupCommand extends Command {
 
                 const confirmed = await this.ui.confirm(`Do you wish to set up ${name}?`, true);
                 if (!confirmed) {
+                    if (step.onUserSkip) {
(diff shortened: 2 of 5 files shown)


extensions/systemd/index.js
```diff
@@ -22,6 +22,9 @@ class SystemdExtension extends Extension {
                 }
 
                 return false;
+            },
+            userSkip: ({instance}) => {
```
acburdine, line 26: I think this approach makes sense - though I wonder if it might be better named, like `onSkip` or something 🤔

vikaspotluri123: My issue with `onSkip` is it's ambiguous as to what skipped (auto or user) - thoughts on `onUserSkip`?

acburdine: yeah that sounds good 👍

lib/commands/setup.js
```diff
@@ -141,6 +141,10 @@ class SetupCommand extends Command {
 
                 const confirmed = await this.ui.confirm(`Do you wish to set up ${name}?`, true);
                 if (!confirmed) {
+                    if (step.onUserSkip) {
+                        step.onUserSkip(ctx);
```
acburdine, line 145: do we want to add an await here just to support any promise-based skips?

vikaspotluri123: Sure

extensions/systemd/index.js
```diff
@@ -22,6 +22,9 @@ class SystemdExtension extends Extension {
                 }
 
                 return false;
+            },
+            onUserSkip: ({instance}) => {
+                instance.config.set('process', 'local').save();
```
acburdine, line 27: last thing - do we want to print out a log message saying that we're reverting back to local process manager?

vikaspotluri123: Ooh yeah, good idea - `Using "local" process manager` (yellow)?

acburdine: maybe "Systemd setup skipped, reverting to local process manager", yellow sounds good color-wise though

acburdine (approved):

acburdine:
@vikaspotluri123 looks like the log isn't being supplied for the test

acburdine:
also had another thought, would it make sense to add a check to the start of systemd setup to switch the process to systemd if it's not set to that already?

thinking if someone skips the check the first time around, if they run `ghost setup systemd` later we'd likely want to change their process manager back to systemd.

extensions/systemd/index.js
```diff
@@ -22,18 +22,28 @@ class SystemdExtension extends Extension {
...
-    _setup({instance}, task) {
+    _setup({instance, ui}, task) {
         const uid = getUid(instance.dir);
 
         // getUid returns either the uid or null
         if (!uid) {
             return task.skip('The "ghost" user has not been created, try running `ghost setup linux-user` first');
         }
 
+        if (instance.config.get('process') !== 'systemd') {
```
acburdine, line 41: we might need to change the enable function as well, since it checks if process is systemd

vikaspotluri123: What could we do for that? Just `!argv.local`? Or `!argv.local || argv.force` 🤔

acburdine: maybe check if `argv.stages` contains systemd? https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/setup.js#L127-L129

vikaspotluri123: Can't suggest a patch, but were you thinking this?

```js
enabled: ({instance, argv}) => !argv.local && (instance.config.get('process') === 'systemd' || argv.stages.includes('systemd')),
```

Or completely getting rid of the `instance.config` check?

acburdine: probably
```js
enabled: ({instance, argv}) => !argv.local && (instance.config.get('process') === 'systemd' || (argv.stages && argv.stages.includes('systemd'))),
```

acburdine: sidenote, can't wait until optional chaining is available in node LTS 😂

vikaspotluri123: If we switch to typescript we could do it today 😛

Issue #1265, opened 2020-08-01
Support for Ubuntu 20.04


halkyon:
Ubuntu 20.04 is available now, but the stack check when installing Ghost is requiring Ubuntu 16 or 18 specifically.
I've worked around it using `--no-stack` for now.

Doesn't look like any issues with Ubuntu 20.04, but I haven't done an exhaustive check.

acburdine:
@halkyon We've done a bit of looking into Ubuntu 20.04, and the main issue with it at the moment is that, by default, Ubuntu 20.04 installs MySQL 8, instead of MySQL 5.7 (the default for Ubuntu 16 & 18).

Last I checked, there were some compatibility issues between Ghost itself & MySQL 8. Once those compatibility issues are fixed (they may have been fixed already, need to check) and we get any documentation updates done for steps that need to be taken to work on Ubuntu 20, we can look at allowing Ubuntu 20 in the stack check.

ErisDS:
MySQL 8 is in a weird state. There's no official support for it in Node.js yet!

You can read my recent summary of the state of it here: https://github.com/TryGhost/Ghost/issues/11625#issuecomment-644879046

The main things from a Ghost CLI perspective are auth (not sure if it's impacted) and the fact that upgrading from 5 -> 8 will cause mixed collations which is _definitely_ not supported and would never be.

dsecareanu:
Installing ghost on ubuntu 20.04 works using mariadb and having the database credentials setup before hand (though it doesn't help much with the default setup of ubuntu 20.04 and the ghost install command, but at least the 1-click install in DO and other similar options can be update and the install docs could mention this caveat.

IvanMoReau:
> Installing ghost on ubuntu 20.04 works using mariadb and having the database credentials setup before hand (though it doesn't help much with the default setup of ubuntu 20.04 and the ghost install command, but at least the 1-click install in DO and other similar options can be update and the install docs could mention this caveat.

I can actually confirm this.

![Captura de pantalla de 2020-10-02 15-30-26](https://user-images.githubusercontent.com/10843250/94967390-954ec280-04c4-11eb-8188-6156e8f6d38b.png)

acburdine:
Apologies for the delayed response on this - gonna shoot to get basic mysql 8 + ubuntu 20 support out in the next release

acburdine:
[...]

Issue #853, opened 2018-11-06
Check nginx install without dpkg


enguerran:
### Summary

Regarding https://github.com/TryGhost/Ghost-CLI/blob/master/extensions/nginx/index.js#L282
```js
    isSupported() {
        try {
            execa.shellSync('dpkg -l | grep nginx', {stdio: 'ignore'});
            return true;
        } catch (e) {
            return false;
        }
    }
```

We check Nginx by using a third-party command called `dpkg` that is available only on Debian-based distributions.
I use Arch linux and I have nginx installed. I can check it with 

```js
    isSupported() {
        try {
            execa.shellSync('nginx -v', {stdio: 'ignore'});
            return true;
        } catch (e) {
            return false;
        }
    }
```

Then the command `ghost setup nginx` can do everything it needs:

```
/v/w/ghost> ghost setup nginx
✔ Creating nginx config file at /var/www/ghost/system/files/servername.conf
+ sudo ln -sf /var/www/ghost/system/files/servername.conf /etc/nginx/sites-available/servername.conf
+ sudo ln -sf /etc/nginx/sites-available/servername.conf /etc/nginx/sites-enabled/servername.conf
+ sudo nginx -s reload
✔ Setting up Nginx
```

Can we not consider using `nginx -v` to check for the presence of nginx?

enguerran:
After an update, I find this error on my Arch Linux server (which doesn't have dpkg -l to search if nginx is installed).

I thought that maybe one solution was to ask the user if we skip or not after `isSupported` returned `false`. Because in the end, in the following, we consider that `nginx` is in the `PATH` or provided via the environment variable.
https://github.com/TryGhost/Ghost-CLI/blob/master/extensions/nginx/index.js#L275

The solution may be something like this
https://github.com/TryGhost/Ghost-CLI/blob/94389d559700854fb4c1502643d76eafe48a61f2/lib/commands/doctor/checks/system-stack.js#L43-L62

I can write the code needed using inquirer if the proposotion is accepted.

acburdine:
@enguerran I'm working on some improvements to these checks - should have something out soon.

acburdine:
@enguerran I've refactored the stack check to use [systeminformation's](https://systeminformation.io) `services` utility to check if `nginx` is installed by iterating over the list of installed services, rather than distro-specific command like `dpkg`.

enguerran:
[...]

PR #1313, opened 2020-10-09
feat(stack): refactor stack check & allow ubuntu 20

acburdine:
refs #853,#1265
- refactor system stack check to use systeminformation to check support
- update ubuntu allowed versions to support Ubuntu 20


vikaspotluri123 (approved):
LGTM 👍🏾 

We'll probably want to add an extension hook for other process managers (thinking about supervisor) to run checks here

PR #1315, opened 2020-10-09
feat(nginx): refactor to use async/await

acburdine:


extensions/nginx/index.js
```diff
@@ -135,46 +127,45 @@ class NginxExtension extends Extension {
...
         return this.ui.listr([{
             title: 'Checking DNS resolution',
-            task: () => Promise.fromNode(cb => dns.lookup(parsedUrl.hostname, {family: 4}, cb)).catch((error) => {
-                if (error.code !== 'ENOTFOUND') {
-                    // Some other error
-                    return Promise.reject(new CliError({
-                        message: `Error trying to lookup DNS for '${parsedUrl.hostname}'`,
-                        err: error
-                    }));
+            task: async () => {
```
vikaspotluri123, line 130: Is it worth using async / await here? Feels like promise / catch works better in this scenario

acburdine: yeah, guess it doesn't really add too much

extensions/nginx/index.js
```diff
@@ -213,77 +196,81 @@ class NginxExtension extends Extension {
...
-            return Promise.resolve();
+            return;
         }
 
         const parsedUrl = url.parse(instanceUrl);
         const confFile = `${parsedUrl.hostname}.conf`;
         const sslConfFile = `${parsedUrl.hostname}-ssl.conf`;
 
-        const promises = [];
+        let restart = false;
```
vikaspotluri123, line 219: Looks like these commands went from parallel to serial - was that expected?

vikaspotluri123 (approved):

extensions/nginx/index.js
acburdine: yeah - I didn't think the benefit of running them in parallel outweighed code cleanliness. if these were super slow commands then it might make sense to run them in parallel but since these are just file deletions it's fine to run them in series

PR #1317, opened 2020-10-09
feat(ci): add end-to-end github actions tests for standard stack

acburdine:
these will start working when #1318 and #1319 are merged in

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -1,0 +1,102 @@
+name: End-To-End Testing
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  ubuntu-16:
+    runs-on: ubuntu-16.04
+    name: 'Ubuntu 16.04'
+    steps:
+      - run: sudo systemctl start mysql.service && sudo apt-get update && sudo apt-get install -y nginx
+        name: 'Setting up required services'
+      - run: 'echo -e "127.0.0.1 cli-testing.ghost.org\n" | sudo tee -a /etc/hosts'
+        name: 'Updating hosts file'
+      - uses: actions/checkout@v2
+        with:
+          path: 'cli'
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - run: cd cli && yarn && sudo yarn link --link-folder /usr/bin/
+      - name: Setting up Ghost instance
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,106 @@
...
+        name: 'Setting up required services'
+      - run: 'echo -e "127.0.0.1 cli-testing.ghost.org\n" | sudo tee -a /etc/hosts'
+        name: 'Updating hosts file'
+      - uses: actions/checkout@v2
+        with:
+          path: 'cli'
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - run: cd cli && yarn && sudo yarn link --link-folder /bin/
```
vikaspotluri123, line 23: Is there a reason we need to use link? What's wrong with running `./cli/bin/ghost ...`?

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Setting up Ghost instance
+        run: |
+          mkdir ghost
```
vikaspotluri123, line 26: Shouldn't be necessary (implicit test)

vikaspotluri123 (approved):
Couple q's, but everything else looks good. Once this is done, I'll rebase my PR and do a minimal matrix (ubuntu 16, ubuntu 18, ubuntu 20, ubuntu 20 MySQL 8)

.github/workflows/e2e-test.yml
acburdine: that could work too - wanted to try and make it as close to real-world as possible though

.github/workflows/e2e-test.yml
acburdine: good point, will remove 👍

.github/workflows/e2e-test.yml
vikaspotluri123: Gotcha. Main reason I was asking is that seems to be the only difference between the 3 OSes - if they follow the same template, matrixing will be super easy

acburdine: I'll test and see if /usr/bin works on ubuntu 16/18.

acburdine: yup looks like `/usr/bin/` works consistently across all three - updated to just use that

PR #1319, opened 2020-10-09
fix(stack): refactor nginx/systemd checks to use same method

acburdine:


lib/commands/doctor/checks/system-stack.js
```diff
@@ -7,21 +7,10 @@ const taskTitle = 'Checking system compatibility';
...
-    } catch (error) {
-        return false;
-    }
-}
-
-async function hasNginx() {
-    try {
-        const services = await sysinfo.services('*');
-        return services.some(s => s.name === nginxProgramName);
+        const services = await sysinfo.services(name);
```
vikaspotluri123, line 12: Not sure how long this takes; do we want to cache it?

vikaspotluri123 (approved):

lib/commands/doctor/checks/system-stack.js
acburdine: nah, it's a fairly straightforward `ps` call that does some grepping on the output, so it shouldn't take long at all to run.

PR #1320, opened 2020-10-11
feat(extension): rewrite template method

acburdine:
The idea behind this PR is to rethink how we're handling the generation of configuration files for things like nginx + systemd. Currently, the configuration files are stored inside the Ghost site's installation directory and symlinked into the proper directory on the system, but this has come with a few drawbacks, and does not in hindsight provide that much of a benefit.

If a person wants to modify their configuration files after Ghost-CLI generates them, then they can visit the directories on the system that those files normally reside in (i.e. `/etc/nginx` for nginx) and edit them there.

As a side benefit, this removes the one significant complication that would prevent us from converting our usage of `inquirer` to `enquirer`.

### DIFF
--- a/lib/extension.js
+++ b/lib/extension.js
@@ -1,5 +1,6 @@
 // @ts-check
 const fs = require('fs-extra');
+const os = require('os');
 const path = require('path');
 const mapValues = require('lodash/mapValues');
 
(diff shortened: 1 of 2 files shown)


lib/extension.js
```diff
@@ -56,78 +57,24 @@ class Extension {
      * @param {string} contents Contents of the template file to create
      * @param {string} descriptor Description of the template file (used in the prompt)
      * @param {string} file Filename
-     * @param {string} dir Directory to link the config file into
-     * @return {Promise<boolean>} True if the config file was created successfully, otherwise false
+     * @param {string} dir Directory to move the config file into
+     * @return {Promise} True if the config file was created successfully, otherwise false
```
vikaspotluri123, line 61:
```suggestion
     * @return {Promise<void>}
```
Looks like it doesn't return a boolean anymore, so the description is no longer accurate 😅

vikaspotluri123 (approved):
Overall, the changes make sense. I think the only benefit for symlinking is knowing what files were created by the CLI outside of the install directory, but we can continue work on that with the transparency file in the future.

 I'm guessing we're not logging the final output path since running sudo logs the move command?

vikaspotluri123:
As an aside, do we need the acceptance tests now that we have the E2E tests?

lib/extension.js
acburdine: done 👍

acburdine:
> I'm guessing we're not logging the final output path since running sudo logs the move command?

yeah, I figure the `mv` command is probably informative enough

PR #1321, opened 2020-10-11
feat(ci): rework test setup

acburdine:


vikaspotluri123 (approved):
Windows grep be crazy 😆 Ended up with the same number of jobs, but cumulative time is probably lower 😎

PR #1332, opened 2020-10-29
feat(install): added `canary` as an install target

daniellockyer:
- We want to allow Ghost-CLI to download canary builds of the latest commits in Ghost & Ghost-Admin to allow people to test new changes
- This commit allows us to do things like `ghost install canary --local`, which will download the latest CI artifact zip from GitHub, and install it like usual.
- Due to a limitation in the GitHub API, the download link requires a GitHub auth token. This has been implemented as reading it from the GST_TOKEN environment variable, but hopefully we have a workaround for that in the future.

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -12,7 +12,13 @@ class InstallCommand extends Command {
         return yargs;
     }
 
-    run(argv) {
+    async run(argv) {
+        const fs = require('fs-extra');
+        const os = require('os');
+        const decompress = require('decompress');
+        const download = require('download');
+        const got = require('got');
+
         const errors = require('../errors');
         const yarnInstall = require('../tasks/yarn-install');
         const dirIsEmpty = require('../utils/dir-is-empty');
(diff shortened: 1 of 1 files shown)


daniellockyer:
@acburdine I've got a couple of questions, however none are urgent. Just wanted to see if you had an immediate fix:

* My aim is to use this with prereleases, which are built in CI, and have versions like `3.37.0-pre.d86dca244.f637be76b.0`. However, Ghost-CLI shows `Downloading and installing Ghost v3.37.0` during the install because [this line](https://github.com/TryGhost/Ghost-CLI/blob/e46019175ad6bc8c2e90c8cba68e81d916149a9a/lib/utils/version.js#L49) strips the prerelease version. Is there an easy/*"non-breaking for the rest of CLI"* way around this?
* Because I inserted this code in before the normal UI starts, there are a few seconds when it's downloading the zip where there's no output. I don't think it's a big deal, but is there an easy way to output some messages that match the UI output?

acburdine:
@daniellockyer 
> My aim is to use this with prereleases, which are built in CI, and have versions like 3.37.0-pre.d86dca244.f637be76b.0. However, Ghost-CLI shows Downloading and installing Ghost v3.37.0 during the install because this line strips the prerelease version. Is there an easy/"non-breaking for the rest of CLI" way around this?

There probably is - it might involve moving this code to a better location in the codebase though 🤔 I'll do some thinking about where might be best to put this.

> Because I inserted this code in before the normal UI starts, there are a few seconds when it's downloading the zip where there's no output. I don't think it's a big deal, but is there an easy way to output some messages that match the UI output?

Yup! the UI class (available as `this.ui` in the install command) has a `run` method which outputs a spinner while an async task is running: https://github.com/TryGhost/Ghost-CLI/blob/master/lib/ui/index.js#L80-L91

acburdine:
probably worth hooking this up to the new `--channel` flag being added in #1390

Issue #1338, opened 2020-11-10
Display GitHub release notes when installing/updating Ghost


daniellockyer:
### Problem

Ghost-CLI doesn't currently display information to the user about releases. It'd be nice to show the description of a Release from GitHub (like https://github.com/TryGhost/Ghost/releases/tag/3.37.1) when the user is installing/updating, to let them know about the contents of the release and potential issues.

We usually use the ⚠️ in release notes when there's a possibility of data loss or loss of critical functionality. There have been times where there's an important bug in the most recent version and we've added it to the release notes (https://github.com/TryGhost/Ghost/releases/tag/3.35.0), but we can't be sure anyone actually reads them before doing a `ghost update`.

### Solution

Pull down https://api.github.com/repos/TryGhost/Ghost/releases, and display the one relevant to the release we're installing/updating to to the user as a step in the process.

Further ideas:
* If the notes contain a ⚠️, maybe re-confirm they'd like to proceed (?).

PR #1340, opened 2020-11-16
feat(update): display release notes when updating Ghost

daniellockyer:
closes https://github.com/TryGhost/Ghost-CLI/issues/1338

- we want to be able to display the release notes so we can inform users
  about new features

### DIFF
--- a/lib/tasks/release-notes.js
+++ b/lib/tasks/release-notes.js
@@ -1,0 +1,21 @@
+module.exports = async (ctx, task) => {
+    const got = require('got');
+    let response;
+
+    try {
+        response = await got('https://api.github.com/repos/TryGhost/Ghost/releases', {json: true});
+    } catch (err) {
+        task.title = 'Unable to fetch release notes';
+        return;
+    }
+
+    const relevantNotes = response.body.filter(note => note.tag_name === ctx.version)[0];
+
+    if (!relevantNotes) {
+        task.title = 'Release notes were not found';
+        return;
+    }
+
+    task.title = 'Fetched release notes';
+    ctx.ui.log(`\n${relevantNotes.body}\n`, 'green');
+};
--- a/lib/commands/update.js
+++ b/lib/commands/update.js
@@ -21,6 +21,7 @@ class UpdateCommand extends Command {
         const MigrateCommand = require('./migrate');
         const migrator = require('../tasks/migrator');
         const majorUpdate = require('../tasks/major-update');
+        const releaseNotes = require('../tasks/release-notes');
 
         const instance = this.system.getInstance();
 
@@ -70,6 +71,9 @@ class UpdateCommand extends Command {
 
         // TODO: add meaningful update checks after this task
         const tasks = [{
+            title: 'Fetching release notes',
+            task: releaseNotes
+        }, {
             title: 'Downloading and updating Ghost',
             skip: ({rollback}) => rollback,
             task: this.downloadAndUpdate
--- a/test/unit/commands/update-spec.js
+++ b/test/unit/commands/update-spec.js
@@ -191,11 +191,13 @@ describe('Unit: Commands > Update', function () {
                 rollback: sinon.stub().resolves()
             };
 
+            const releaseNotesStub = sinon.stub().resolves();
             const majorUpdateStub = sinon.stub();
 
             const UpdateCommand = proxyquire(modulePath, {
                 '../tasks/migrator': migratorStub,
-                '../tasks/major-update': majorUpdateStub
+                '../tasks/major-update': majorUpdateStub,
+                '../tasks/release-notes': releaseNotesStub
             });
 
             const ghostConfig = configStub();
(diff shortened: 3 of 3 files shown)


lib/tasks/release-notes.js
```diff
@@ -0,0 +1,27 @@
...
+    const {CliError} = require('../errors');
+
+    const question = 'The release notes contain a ⚠️, which indicates an issue with the release. Are you sure you want to continue?';
+    const answer = await ui.confirm(question, true);
+
+    if (answer) {
+        return Promise.resolve();
+    }
+
+    throw new CliError('');
```
daniellockyer, line 26: WIP

lib/tasks/release-notes.js
```diff
@@ -0,0 +1,27 @@
...
+    const relevantNotes = response.body.filter(note => note.tag_name === ctx.version)[0];
+
+    task.title = 'Fetched release notes';
+
+    if (!relevantNotes) {
+        return;
+    }
+
+    ctx.releaseNotesContainWarning = relevantNotes.body.includes('⚠️');
+    task.output = '\n' + relevantNotes.body + '\n';
```
acburdine, line 13: you shouldn't need to do this & modify the task renderer, you should be able to just log things via the UI log method and it'll output correctly

daniellockyer: Good point - fixed in temporary new commit

lib/tasks/release-notes.js
acburdine: probably this, just give it an informative error message

lib/tasks/release-notes.js
```diff
@@ -0,0 +1,27 @@
...
+};
+
+module.exports.checkForConfirmation = async ({ui}) => {
+    const {CliError} = require('../errors');
+
+    const question = 'The release notes contain a ⚠️, which indicates an issue with the release. Are you sure you want to continue?';
+    const answer = await ui.confirm(question, true);
+
+    if (answer) {
+        return Promise.resolve();
```
acburdine, line 23:
```suggestion
        return;
```

lib/tasks/release-notes.js
```diff
@@ -0,0 +1,29 @@
...
+
+    ctx.releaseNotesContainWarning = relevantNotes.body.includes('⚠️');
+    ctx.ui.log('\n' + relevantNotes.body + '\n', 'green');
+};
+
+module.exports.checkForConfirmation = async (ctx) => {
+    const {CliError} = require('../errors');
+
+    const question = 'The release notes contain a ⚠️, which indicates an issue with the release. Are you sure you want to continue?';
+    const answer = await ctx.ui.confirm(question, true);
```
daniellockyer, line 20: @acburdine What are your thoughts on the default here?

acburdine: I think true makes sense as default

lib/commands/install.js
```diff
@@ -44,6 +45,14 @@ class InstallCommand extends Command {
         }, argv, {local})).then(() => this.ui.listr([{
             title: 'Checking for latest Ghost version',
             task: this.version
+        }, {
+            title: 'Fetching release notes',
+            task: releaseNotes.displayReleaseNotes,
+            enabled: () => process.env.GHOST_SKIP_RELEASE_NOTES !== 'true'
```
daniellockyer, line 51: @acburdine I've added this get-out clause so we can skip over this feature in the event GitHub are having issues.

daniellockyer:
@acburdine Aside from your quick thoughts on https://github.com/TryGhost/Ghost-CLI/pull/1340#discussion_r526830094, I'm happy with this and would be good to merge 🙂

lib/commands/install.js
```diff
@@ -44,6 +45,9 @@ class InstallCommand extends Command {
         }, argv, {local})).then(() => this.ui.listr([{
             title: 'Checking for latest Ghost version',
             task: this.version
+        }, {
```
acburdine, line 48: does it make sense to show release notes on install? this may add a lot of unnecessary output for someone who's just getting set up for the first time

lib/tasks/release-notes.js
```diff
@@ -0,0 +1,21 @@
...
+
+    const relevantNotes = response.body.filter(note => note.tag_name === ctx.version)[0];
+
+    if (!relevantNotes) {
+        task.title = 'Release notes were not found';
+        return;
+    }
+
+    task.title = 'Fetched release notes';
+    ctx.ui.log('\n' + relevantNotes.body + '\n', 'green');
```
acburdine, line 20: incredibly minor nitpick for consistency: 
```suggestion
    ctx.ui.log(`\n${relevantNotes.body}\n`, 'green');
```

lib/commands/install.js
daniellockyer: True! Removing...

acburdine (approved):

lib/tasks/release-notes.js
```diff
@@ -0,0 +1,21 @@
+module.exports = async (ctx, task) => {
+    const got = require('got');
+    let response;
+
+    try {
+        response = await got('https://api.github.com/repos/TryGhost/Ghost/releases', {json: true});
```
vikaspotluri123, line 6: Was going to mention this before it got merged, but I think we should have a timeout here (between 500ms and 1.5s) since by default there isn't one

acburdine: that's a good point, yeah

PR #1390, opened 2021-01-28
feat(channel): add ability to specify release channel for local installs

acburdine:
this adds a new `--channel` flag to the install command that will allow installing prereleases for local installs if desired.


daniellockyer (approved):
LGTM!

PR #1411, opened 2021-02-25
feat(local): handle v4 boot events in local process manager

acburdine:
no issue:
- Ghost in v4+ emits two events on successful start: 'started' and 'ready', so we need to handle both
- The error message structure in v4 events also changes, so we need to handle things there as well


ErisDS:
> The error message structure in v4 events also changes, so we need to handle things there as well

It does?! Can you elaborate maybe I borked something....

acburdine:
> It does?! Can you elaborate maybe I borked something....
It seemed like it changed based on how the CLI code is currently structured.....that said the code in Ghost v3 looks like it's doing the same thing v4 is doing wrt how errors are being sent 🤔 

Could be the CLI was doing something wrong before, will do some more testing

PR #1412, opened 2021-02-25
feat(boot): handle v4 boot events in port-polling util

acburdine:
no issue
- handles separate 'ready' event from Ghost v4+


vikaspotluri123 (approved):

acburdine:
note: this requires https://github.com/TryGhost/Utils/pull/136 to be in place to fully work, but for the moment I'm going to merge since this works with that change in place.

PR #1417, opened 2021-03-01
 Add support for more environments than production and development

vikaspotluri123:
Does a few things:

1. Updates default environment logic
  a. Adds `--environment (-e)` flag to specify the environment (still has explicit handling for `--development`)
```javascript
/**
 * From user input (overrides operating system default):
 * 1. If the user passes --development or -d, use development
 * 2. If the user passes --environment, use whatever they provided
 * 3. Pull from NODE_ENV
 *
 * From Operating System:
 * 1. Linux is the only officially supported production environment, so default to production
 * 2. All other Operating Systems are assumed to be development
*/
```

2. `system.setEnvironment` now takes in a string
   a. I removed `system.production` and `system.development` since I couldn't find any more references (it was used by `instance.checkEnvironment` but I refactored that too). **Do we want to add a getter with a warning for these**? In case I missed something, this will prevent weird edge cases in the wild since both will be undefined

3. `ls` had a race condition, so I switched from `map` to `mapSeries`.
  a. System is a service, and checking if an instance is running is async. When running these in parallel, each instance will mutate the System environment, which causes the discovered environment of some instances to be incorrect
  b. We might want to revisit this separately. I only saw this when e2e testing
4. Refactor `instance.isRunning` to loop over all `config.*.json` environments (order is 'production', 'development', whatever). Previously only looked at development and production
5. Refactor `instance.checkEnvironment` to throw an error when the environment config doesn't exist. Keeps the production -> development env switch though. 
6. Update some types!


lib/commands/uninstall.js
```diff
@@ -35,8 +36,8 @@ class UninstallCommand extends Command {
             task: () => this.ui.sudo(`rm -rf ${path.join(instance.dir, 'content')}`)
         }, {
             title: 'Removing related configuration',
-            task: () => {
-                this.system.setEnvironment(!fs.existsSync(path.join(instance.dir, 'config.production.json')));
+            task: (_, context) => {
+                this.system.setEnvironment(context._environment);
```
vikaspotluri123, line 40: I'm not sure how much we need to do this store and fetch dance. Do you know why it was implemented like this previously?

lib/utils/discover-envs.js
```diff
@@ -0,0 +1,36 @@
...
+    for (const entry of directoryContents) {
+        const matchResult = environmentExtractor.exec(entry);
+
+        if (matchResult) {
+            environments.push(matchResult[1]);
+        }
+    }
+
+    // Always return in order: production, development, anything else
+    let indexPointer = environments.indexOf('development');
```
acburdine, line 23: would it be easier/more readable to use `Array.sort` here? something like https://runkit.com/acburdine/603cc993f37dbb001a14bdbe might work

acburdine (commented):
LGTM code-wise other than a couple of small comments, will do some more rigorous manual testing though as well

lib/utils/discover-envs.js
vikaspotluri123: Done, I had 3-4 implementations that didn't work so I eventually decided to brute force it 😅

vikaspotluri123:
I think one edge case we have to consider is uninstall since `instance.isRunning()` will set the environment to the first running instance. Super edge case so I'm not worried

acburdine:
@vikaspotluri123 looks like there's a failing test (maybe due to some changes I made in main?) if you can fix the test issues, I think this PR's good to merge

vikaspotluri123:
I found an issue with:

1. Linux
2. No production config
3. yes development config
4. Tries to run in production

I feel like that should fall back to development.

vikaspotluri123:
Rebasing fixed the issue, it was the same as we were seeing in #1475 

Also updated the base to 1.x 😄

acburdine (approved):

PR #1428, opened 2021-03-16
fix(release-notes): add support for 4.x+ tag names

vikaspotluri123:
Also adds a 'lil perf improvement 💨 w/ tests!

### DIFF
--- a/lib/tasks/release-notes.js
+++ b/lib/tasks/release-notes.js
@@ -9,7 +9,7 @@ module.exports = async (ctx, task) => {
         return;
     }
 
-    const relevantNotes = response.body.filter(note => note.tag_name.replace('v', '') === ctx.version)[0];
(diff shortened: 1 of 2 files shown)


lib/tasks/release-notes.js
```diff
@@ -3,13 +3,13 @@ module.exports = async (ctx, task) => {
     let response;
 
     try {
-        response = await got('https://api.github.com/repos/TryGhost/Ghost/releases', {json: true, timeout: 5000});
+        response = await got.get('https://api.github.com/repos/TryGhost/Ghost/releases', {json: true, timeout: 5000});
```
vikaspotluri123, line 6: Makes stubbing so much easier!

acburdine: got to get those tests 💯

acburdine (approved):

PR #1435, opened 2021-03-18
feat(install): cleanup install directory on error

acburdine:


vikaspotluri123 (approved):
We don't need to have an extension hook because that's handled by setup, right?

acburdine:
> We don't need to have an extension hook because that's handled by setup, right?

yup, none of the extension logic is set up yet because extensions are only called during setup.

PR #1436, opened 2021-03-19
feat(port-polling): log messages coming from Ghost

acburdine:


vikaspotluri123 (approved):
LGTM code-wise

I think we might need to audit what port-polling utils we support now that 1.x and 2.x are EoL

acburdine:
> I think we might need to audit what port-polling utils we support now that 1.x and 2.x are EoL

Yeah I planned to do some of that but figured for this go-round I'd keep the changes small 😅

PR #1475, opened 2021-05-22
fix(start): fix warning behavior

acburdine:


lib/commands/start.js
```diff
@@ -32,16 +32,15 @@ class StartCommand extends Command {
             return;
         }
 
-        const url = instance.config.get('url');
+        instance.checkEnvironment();
 
-        if (this.system.environment === 'production' && url.startsWith('http://')) {
+        if (this.system.environment === 'production' && instance.config.get('url', '').startsWith('http://')) {
```
vikaspotluri123, line 37: Related question - should Ghost start if there isn't a URL? 😄

acburdine: probably not - though I feel like that's probably a check for a different place in the code 😅

ErisDS:
No `url` should be an invalid config but agree that can be dealt with elsewhere.

Interestingly, only the production Ghost config doesn't have a default URL... maybe it too should be localhost:2368 like in development 🤔 but this at least explains why it's a fairly rare occurence. 

Not sure what the wider ramifications are here...

PR #1494, opened 2021-06-18
Migrate to lerna monorepo

acburdine:
This PR sets up the repository as a monorepo, and moves all of the existing CLI code into the only package (for now) called ghost-cli.

Eventually, code will be split out into separate packages, but for now this PR attempts to do as few things as possible to get the monorepo to work.


vikaspotluri123 (approved):

lerna.json
```diff
@@ -0,0 +1,16 @@
+{
+  "version": "independent",
+  "npmClient": "yarn",
+  "packages": ["packages/*"],
+  "useWorkspaces": true,
+  "command": {
+    "publish": {
+      "allowBranch": "main"
```
ErisDS, line 8: You have the same config as other repos except you're missing:

`"message": "Published new versions"`

Without this the commit message will just be "Publish". I would have thought for CLI you would at least want to update it to use the semantic stuff i.e. chore: Publish?

jsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "include": ["lib/**/*.js", "extensions/**/*.js"],
+  "include": ["packages/**/*.js"],
   "compilerOptions": {
     "checkJs": true,
     "module": "commonjs",
```
ErisDS, line 5: Just flagging as I see it that you have es2019 in here, and 2018 in eslint.

The rest of the Ghost ecosystem is currently targetting 2018, although I have no strong feels about this.

Also does `moduleResolution: node` mean you don't need this:

``` 
 "exclude": [
    "node_modules"
  ]
```

(I'm comparing against the Utils repo)

packages/ghost-cli/package.json
```diff
@@ -28,7 +28,7 @@
   "scripts": {
     "coverage": "nyc report --reporter=text-lcov | coveralls",
     "lint": "eslint . --ext js --cache",
-    "test": "nyc --reporter=html --reporter=text mocha -t 5000 --recursive test/unit extensions/**/test",
+    "test": "nyc --reporter=html --reporter=text mocha --color -t 5000 --recursive test/unit extensions/**/test",
```
ErisDS, line 31: Really interesting!

package.json
```diff
@@ -0,0 +1,18 @@
...
+  "repository": "https://github.com/TryGhost/Ghost-CLI",
+  "author": "Ghost Foundation",
+  "license": "MIT",
+  "workspaces": ["packages/*"],
+  "eslintignore": ["**/node_modules/**"],
+  "scripts": {
+    "presetup": "yarn",
+    "setup": "lerna bootstrap",
+    "test": "lerna run test",
+    "lint": "lerna run lint"
```
ErisDS, line 12: Missing a ship command?

lerna.json
```diff
@@ -0,0 +1,16 @@
...
+    }
+  },
+  "local": {
+    "public": true,
+    "repo": "https://github.com/TryGhost/Ghost-CLI",
+    "scope": "@tryghost"
```
ErisDS, line 14: Not sure if this will affect how the ghost-cli package gets published 🤔  hopefully it only affects new packages?

ErisDS (commented):
A couple of other notes that don't necessarily need anything right now, just whilst I see them:


- Everywhere else we've switched to eslintrs.js rather than  .json. The eslint.json in this project has an error because 2018 is in quotes and should be a number. Swapping to .js is highly recommended.
- There are no top-level tests here. To be fair, we don't use those in most other repos, but this is the one where I could see that potentially being useful? Just a thought
- As you're moving to publishing multiple packages in the `@tryghost` scope you'll need to be wary of naming. I recommend establishing a convention as we have in some repos of always using a prefix. So any packages would be e.g. `@tryghost/cli-something`.

Only thing that really needs a sense check before merging is what the workflow for publishing will be. The existing ship command is currently in the packages/ghost-cli package.json stil, and there's no global command. Need to ensure we end up with `npm i ghost-cli` and not `npm i @tryghost/ghost-cli` but the other packages end up being `@tryghost/cli-something`.

lerna.json
acburdine: yeah - was gonna look into that as part of the release workflow

lerna.json
acburdine: from the looks of it this `local` config is only used by slimer-cli, which seems to only be for new packages?

package.json
acburdine: yup, was gonna tackle that as part of the release workflow as well

jsconfig.json
acburdine: forgot to update the eslint version 😬 

Reason I put 2019 here is that 2019 is the newest version supported by node 12 (the min version of node 12 we're on has 100% of es2019 features implemented) - figured since we're doing a decent rewrite anyways it would make sense to use as new features as we could

acburdine: > Also does moduleResolution: node mean you don't need this:

to be honest I'm not sure `exclude: node_modules` is even needed 🤔 Given that jsconfig.json is essentially the same as tsconfig.json, but without the enforcement/compilation steps, I'm not sure the exclude functionality really does anything, unless you ended up opening a file from node_modules in your editor.

acburdine:
> Everywhere else we've switched to eslintrs.js rather than .json. The eslint.json in this project has an error because 2018 is in quotes and should be a number. Swapping to .js is highly recommended.

this is fixed now

> There are no top-level tests here. To be fair, we don't use those in most other repos, but this is the one where I could see that potentially being useful? Just a thought

It seems like the top-level tests for this repo would be the end-to-end tests that we have in github actions? not sure what the top level node tests would really test, unless it's something to do with a release workflow

> As you're moving to publishing multiple packages in the @tryghost scope you'll need to be wary of naming. I recommend establishing a convention as we have in some repos of always using a prefix. So any packages would be e.g. @tryghost/cli-something

Yup that was the thought - any package other than the main `ghost-cli` one would be `@tryghost/cli-something`

acburdine:
> Only thing that really needs a sense check before merging is what the workflow for publishing will be. The existing ship command is currently in the packages/ghost-cli package.json stil, and there's no global command. Need to ensure we end up with npm i ghost-cli and not npm i @tryghost/ghost-cli but the other packages end up being @tryghost/cli-something.

I'm 95% sure that lerna uses the package name defined in each individual package's `package.json`, so I _think_ we shouldn't have to worry about ghost-cli getting auto-prefixed with `@tryghost`

lerna.json
acburdine: added a basic release workflow 👍🏻

PR #1570, opened 2021-10-17
Correct ghostDeprecated message text

routevegetable:
Now matches the actual deprecated version 3.0.0


acburdine:
@routevegetable Ghost 3.x is supported until Jan 2022 (see [this doc](https://ghost.org/docs/faq/major-versions-lts/)). 

The logic to show this message is currently "if ghost version < 3.0.0, show deprecation message" so the current message matches the logic.

PR #1581, opened 2021-11-19
Added 3 new folders to content

ErisDS:
- Recently we've added 3 new folders to content in Ghost: media, files and public
- This updates Ghost-CLI to create the folders for both new and existing installs

This is not urgent, but the migration before version will need updating in line with whenever this gets released, I think. Not sure I understood how that works.

### DIFF
--- a/packages/ghost-cli/lib/migrations.js
+++ b/packages/ghost-cli/lib/migrations.js
@@ -1,6 +1,5 @@
 const path = require('path');
 
-
 async function ensureFolder(context, folderName) {
     const ghostUser = require('./utils/use-ghost-user');
 
@@ -15,7 +14,7 @@ async function ensureFolder(context, folderName) {
 }
 
 async function ensureSettingsFolder(context) {
-    await ensureFolder(context, 'settings')
+    await ensureFolder(context, 'settings');
 }
 
 async function makeSqliteAbsolute({instance}) {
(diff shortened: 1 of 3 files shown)


packages/ghost-cli/test/unit/migrations-spec.js
```diff
@@ -81,4 +81,54 @@ describe('Unit: Migrations', function () {
...
+            return migrations[0].task(context).then(() => {
+                expect(ghostUserStub.calledThrice).to.be.true;
+                expect(ghostUserStub.calledWithExactly('/var/www/ghost/content')).to.be.true;
+                expect(fsStub.calledThrice).to.be.true;
+                expect(fsStub.firstCall.calledWithExactly('/var/www/ghost/content/media')).to.be.true;
+                expect(fsStub.secondCall.calledWithExactly('/var/www/ghost/content/files')).to.be.true;
+                expect(fsStub.thirdCall.calledWithExactly('/var/www/ghost/content/public')).to.be.true;
+            });
+        });
+    })
```
acburdine, line 133:
```suggestion
    });
```

packages/ghost-cli/lib/migrations.js
```diff
@@ -34,4 +45,8 @@ module.exports = [{
     before: '1.14.1',
     title: 'Fix Sqlite DB path',
     task: makeSqliteAbsolute
+}, {
+    before: '1.17.4',
```
acburdine, line 49: bumping the version here so it runs on instances created with CLI 1.18.0 or earlier (this change will be released as v1.18.0)

```suggestion
    before: '1.18.1',
```

acburdine (commented):
small fixes

PR #1591, opened 2022-01-13
fix(install): remove non-existent flags

Paladiamors:
The following error occurs during installation of ghost using the `ghost install local` command due to the `--no-emoji` and `--no-progress` no longer existing in yarn. Removal of these flags allows installation to continue. This error is also noted in the ghost forum: https://forum.ghost.org/t/yarn-error-when-i-install-ghost-local/27828/3

```
$ ghost install local
✔ Checking system Node.js version - found v16.13.2
✔ Checking memory availability
✔ Checking free space
✔ Checking for latest Ghost version
✔ Setting up install directory
✖ Downloading and installing Ghost v4.32.2
A ProcessError occurred.

Message: Command failed: yarn install --no-emoji --no-progress
```


vikaspotluri123:
I'm on node 16.13.2, and `yarn install -h` shows `--no-progress` as a valid flag, and `--emoji` as a valid flag.

If you're able to reproduce the error using the last released cli on your machine, can you try testing your branch with

`args = ['install', '--emoji', 'false', '--no-progress'];`

and see if the install still fails?

Paladiamors:
@vikaspotluri123 

Sure, I updated the args and reran `ghost install local` with the flags mentioned `const args = ['install', '--emoji', 'false', '--no-progress']`:

```
$ ghost install local

✔ Checking system Node.js version - found v16.13.2
✔ Checking memory availability
✔ Checking free space
✔ Checking for latest Ghost version
✔ Setting up install directory
✖ Downloading and installing Ghost v4.32.2
A ProcessError occurred.

Message: Command failed: yarn install --emoji false --no-progress

Unknown Syntax Error: Unsupported option name ("--emoji").

$ yarn install [--json] [--immutable] [--immutable-cache] [--check-cache] [--inline-builds] [--mode #0]

Exit code: 1


Debug Information:
    OS: Ubuntu, v20.04.3 LTS
    Node Version: v16.13.2
    Ghost-CLI Version: 1.18.1
    Environment: development
    Command: 'ghost install local'
```

Still getting the same error. Does installation work in your environment, or we we some how pulling different versions of yarn?

Paladiamors:
I've taken a look at the installed version of yarn on my side for node and see it's 3.1.1

```
~/.nvm/versions/node/v16.13.2/lib/node_modules/ghost-cli/node_modules/yarn/bin$ ./yarn -v
3.1.1
```

And the flags I get are listed below from a `yarn install -h`:

```
━━━ Usage ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

$ yarn install

━━━ Options ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  --json               Format the output as an NDJSON stream
  --immutable          Abort with an error exit code if the lockfile was to be modified
  --immutable-cache    Abort with an error exit code if the cache folder was to be modified
  --check-cache        Always refetch the packages and ensure that their checksums are consistent
  --inline-builds      Verbosely print the output of the build steps of dependencies
  --mode #0            Change what artifacts installs generate
```

vikaspotluri123:
That seems to be the issue - What version do you get if you run `node ./yarn.js -v`?

Paladiamors:
@vikaspotluri123 The version I'm getting is `3.1.1`

vikaspotluri123:
Hmm, trying in both my dev Ghost CLI, and globally installed version, but both places correctly report `./yarn -v` as  `1.22.15` (which is what the CLI asks for).

I have a feeling this has something to do with yarn 3, which I'll need some time to look into / reproduce. In the meantime, are you able to use yarn 1.x when installing Ghost and 3.x everywhere else?

Paladiamors:
Indeed, I've taken a look at the required version for yarn in the package.json file and it seems to be pointing to 1.22.15, but it's not installing that version for me. my system yarn version is 3.1.1 and that seems to be taking prescidence over the version that is getting installed. I've tried `npm install yarn@1.22.15` but that doesn't seem to be installing it to the `node_modules` directory.

vikaspotluri123:
I think the yarn version that the CLI has is still 1.x, but the yarn binary in the CLI folder is picking up your global yarn install over the local one - can you confirm this by checking the version of `~/.nvm/versions/node/v16.13.2/lib/node_modules/ghost-cli/node_modules/yarn/package.json`?

Paladiamors:
Catting out the file seems to have yarn set to 1.22.5 in there

```
cat ~/.nvm/versions/node/v16.13.2/lib/node_modules/ghost-cli/node_modules/yarn/package.json
{
  "name": "yarn",
  "installationMethod": "tar",
  "version": "1.22.15",
  "license": "BSD-2-Clause",
  "preferGlobal": true,
  "description": "📦🐈 Fast, reliable, and secure dependency management.",
  "resolutions": {
    "sshpk": "^1.14.2"
  },
  "engines": {
    "node": ">=4.0.0"
  },
  "repository": "yarnpkg/yarn",
  "bin": {
    "yarn": "./bin/yarn.js",
    "yarnpkg": "./bin/yarn.js"
  },
  "scripts": {
    "preinstall": ":; (node ./preinstall.js > /dev/null 2>&1 || true)"
  },
  "config": {
    "commitizen": {
      "path": "./node_modules/cz-conventional-changelog"
    }
  }
}
```

oddly enough the installed package in the node directory is:
```
~/.nvm/versions/node/v16.13.2/lib/node_modules/ghost-cli/node_modules/yarn/bin$ ./yarn --version
3.1.1
```

In the end it might not be a problem with the ghost-cli package as the version defined seems correct and the wrong version is being installed for me.

vikaspotluri123:
I think you're right, but I'm going to try to reproduce the issue with yarn 3.x when I get some time - if everyone that uses yarn 3.x runs into this issue, we should do something about it, but if it's a one-off error, we might not 😄

Paladiamors:
Much appreciated if that is possible. As part of the migration process I think the procedure got me to set some variables that makes yarn 3 the default for me which leads to this problem. I haven't found a way to unset it and sometimes the new version of yarn documentation is not terribly informative in terms of what is going on (else I'm just not experienced with it yet).

acburdine:
@Paladiamors sorry for the lack of response on my part

this issue has tended to crop up more and more recently - however unfortunately the fix isn't as simple as removing the flags 😞

For those using yarn 1.x, those flags make it so the CLI output is understandable and isn't cluttered up with yarn progress notifications or emojis. However, the bigger issue is that if the CLI is inadvertently using yarn 3.x to try to install dependencies, there may be other unintended consequences, such as Ghost itself becoming non-functional. I know yarn v2 and v3 default to using pnp modules, which I don't believe has been verified to work with Ghost yet.

Paladiamors:
@acburdine Thanks for responding here. Understandable that there are risks involved with the changes. Also it's fair to say that the version defined by `package.json` should take prescience and that should be the expectation of what runs. Sadly, that setting isn't being respected when running leading to this problem -- probably something that needs to be sorted out by the team that sorts out yarn.

For `--no-emoji` and the `--no-progress` flags, is this only in effect during the installation or whenever yarn is run? If it is just the installation, then the risk might be lower. 

As a anecdotal note, by removing the 2 flags for me, the installation process worked fine for me and I was able to get ghost to run fine in my environment and I'm at the moment doing testing with adding additional pages. So far, so good.

One other thing that could be done, is to do a some kind of try / catch if installation with the flags fail or do some version checking on the installation to see how prevalent this issue is and if it solves the problem for users that end up with yarn2/3 with in their installations. 

Just some thoughts.

ErisDS:
I'm going to close this fix in favour of the one that was merged here: https://github.com/TryGhost/Ghost-CLI/pull/1595

There is probably more work to do in terms of adding a try/catch and a better error message if this case comes up. Would love a PR for that 🙂

PR #1592, opened 2022-01-25
fix(version): handled force flag when specifying versions

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/157

- we've had several bugs in Ghost where users would like to go back to a
  previous version, but `ghost update <version>` doesn't let them go
  back to a version smaller than their current version
- `ghost update <version> --rollback` doesn't work for certain users if
  they've freshly installed Ghost
- by handling `--force`, we can let users choose to go back to an
  arbitrary version at their own discretion
- also adds a test to ensure `force` works with the util

### DIFF
--- a/lib/utils/version.js
+++ b/lib/utils/version.js
@@ -88,7 +88,7 @@ const utils = {
             });
         }
 
-        if (activeVersion && semver.lt(parsed, activeVersion)) {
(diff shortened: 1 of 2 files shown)


acburdine (approved):
LGTM 👍🏻

PR #1596, opened 2022-01-29
fix(deprecation): fix error with missing instance version

vikaspotluri123:


acburdine (approved):
meant to approve before merging 🙃 LGTM

PR #1615, opened 2022-03-08
feat(cmd): added notice about joining the Ghost team

daniellockyer:
refs https://github.com/TryGhost/Toolbox/issues/225

- this commit adds a notice about joining the Ghost team to the CLI that
  should run for every command

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -186,6 +186,11 @@ class Command {
                 await preChecks(ui, system);
             }
 
+            ui.log('');
+            ui.log('Love open source? We’re hiring Node.js Engineers to work on Ghost full-time.', 'magentaBright');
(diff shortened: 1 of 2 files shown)


lib/command.js
```diff
@@ -186,6 +186,11 @@ class Command {
                 await preChecks(ui, system);
             }
 
+            ui.log('');
+            ui.log('Love open source? We’re hiring Node.js Engineers to work on Ghost full-time.', 'magentaBright');
```
acburdine, line 190: I don't _think_ it should - but have you checked to make sure this doesn't get output multiple times for commands that call other commands? (i.e. the `install` command runs `doctor` and `setup` under the hood)

acburdine: yeah that's what I figured, just wanted to be sure 🙂

PR #1639, opened 2022-05-18
Updated backup to export members as well

ErisDS:
- this changes adds a members.csv file if the endpoint exists

Not added any new tests yet 😬


lib/tasks/import/tasks.js
```diff
@@ -40,7 +40,18 @@ async function importTask(ui, instance, exportFile) {
     }], false);
 }
 
-async function exportTask(ui, instance, exportFile) {
+async function tryMembersDownload(version, url, authData, membersFile) {
+    try {
+        await downloadMembersExport(version, url, authData, membersFile);
+    } catch (error) {
+        // Members endpoint may not exist, we can ignore this
+        if (!error.statusCode === 404) {
```
daniellockyer, line 48:
```suggestion
        if (error.statusCode !== 404) {
```

lib/tasks/backup.js
```diff
@@ -8,9 +8,14 @@ const {ProcessError} = require('../errors');
...
 
 module.exports = async function (ui, instance) {
-    // First we need to export the content into a JSON file
-    const contentExportPath = 'data/backup.json';
-    await exportTask(ui, instance, path.join(instance.dir, 'content/', contentExportPath));
+    const datetime = require('moment')().format('YYYY-MM-DD-HH-mm-ss');
+    const backupSuffix = `from-v${instance.version}-on-${datetime}`;
+
+    // First we need to export the content into a JSON file & members into a CSV file
+    const contentExportPath = `data/backup-${backupSuffix}.json`;
```
daniellockyer, line 15:
```suggestion
    const contentExportPath = `data/content-${backupSuffix}.json`;
```

ErisDS: Good point 👍

Issue #1661, opened 2022-07-05
Trust admins to upload large files


markstos:
Currently, admins are not allowed to upload files larger than 50 MB without manually modifing their Nginx config by hand to modify the `client_max_body_size` directive. To support the self-hosting indieweb movement, Ghost should support content creators who wish to support videos and other content larger than 50 MB, as long as this change doesn't compromise security or stablity. 

## History of Ghost's use of Nginx's `client_max_body_size` directive

The current value of `50mb` was added in 2017 to support larger image and theme upload sizes via #231. At the time, there was no discussion of video. There was also no discussion security implications for tuning this value, nor for the possibility of having different values for different `location` blocks. 

## What's the ideal value for `client_max_body_size` ?

Ideally, this directive is to be slightly larger than the max values you'll need. If it's set too high, it can be a vector of denial of service of attacks, allowing malicious users to send a high volume of junk traffic. The tuning of this directive is discussed more here:

  * https://www.acunetix.com/blog/web-security-zone/hardening-nginx/
  * https://www.techrepublic.com/article/how-to-limit-file-upload-size-on-nginx-to-mitigate-dos-attack/

The best way to handle this is to keep the value low for most paths and then only allow high values for specific URLs that are protected by authentication.  For Ghost, most from the public are GET requests with tiny payloads, or simple POST/PUT requests, like signing up as  a member. For all those cases, a value much lower than 50 Mb would work-- 1 Mb would be sufficient.

On the other hand, a trusted admin may wish to upload a 1 GB video for self-hosting. From a security perspective, this adds no significant additional risk-- the admin is already trusted and has access to destroy all the data in the site already, trusting an admin to upload a large file doesn't really add any additional risk.

The biggest "risk" would come from the least trusted users who have the ability to upload photos and videos: "Contributors". 

## Recommended Changes

Nginx allows for a more nuanced approach to setting `client_max_body_size` because it can set in a `location` context as well as the `server` context. Ghost could use that feature to allow unlimited upload sizes to `/ghost/api/admin/media/upload/`, which is only accessible to trusted, authenticated users.  This is the endpoint where videos get uploaded where 50 Mb is too small.  A single `location` block declared using a `regex` syntax could cover all the locations where trusted users could usefully upload larger files.

At the same time, Ghost could consider dialing the 50 mb limit back down to 1mb or some lower value for the majority of endpoints that don't receive file uploads.

ErisDS:
[...]

PR #1662, opened 2022-07-05
feat(nginx): Trust admins to upload large files

markstos:
This request implements the recommendation in https://github.com/TryGhost/Ghost-CLI/issues/1661

However, I've done some more follow-up research that leaves me uncertain on whether to recommend merging this.

## What's checked first: authentication or the upload size?

In the linked issue, I recommended allowing unlimited upload size, but scoped to a route that only works for admins. That's a sound idea, but it only works to block unwanted large uploads if authentication happens before all that data is sent.  So I put together a test:

I submitted an upload to admin-only endpoint as an unauthenticated user, sending a file that was larger than `client_max_body_size`. So my request has two reasons to fail: It's not authenticated and it's too big. The question is: which problem will problem trigger first? 

In my tests, the `client_max_body_size` limit was always triggered before the request was passed to Ghost and the app had a chance to run the authentication middleware.

```
http --print=Hh --multipart https://pea-pod.org/ghost/api/admin/posts/62c44ce15e9a84e9ec9948d4 boom@huge-file.mp4
mp4
POST /ghost/api/admin/posts/62c44ce15e9a84e9ec9948d4 HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 440882259
Content-Type: multipart/form-data; boundary=3a74970ec30842a88d623d35c2a2fa6a
Host: example.com
User-Agent: HTTPie/3.2.1

HTTP/1.1 413 Request Entity Too Large
Connection: close
Content-Length: 192
Content-Type: text/html
Date: Wed, 06 Jul 2022 00:26:09 GMT
Server: nginx/1.18.0 (Ubuntu)
```

To try the force the request to get passed through from Ngnix to Ghost before it was fully uploaded, I tried setting these directives:

```
# In theory, this shoudl stop Nginx from buffering the uploads
proxy_request_buffering off;
# This is supposed to stop Nginx from buffering HTTP/1.1 multi-part uploads
# if the above is set
proxy_http_version 1.1;
```

To summarize: Since Ghost is open source, it doesn't help security to scope large file uploads to a single authenticated route, because that route is known, and Ghost (apparently) can't check the authentication until the upload is complete. And since it's better to be "secure by default", perhaps the current max upload setting of 50 mb is a reasonable default. Those who need to upload videos can change the default.

## Detour: Does `proxy_request_buffering` perform better on or off?

As part testing this, I ended up testing uploading a 400 Mb file repeatedly, both with `proxy_request_buffering on and off. My hypothesis was that uploads would perform better with the buffering off because Ghost could start processing the first part of the file even while the last part is still uploaded.

The result was instead that it was about 7x faster to leave `proxy_request_buffering` enabled. I can't explain why. But here you can see in the Ghost logs the different valuels changes as I toggled the directive back and forth and re-tested:

```
INFO "POST /ghost/api/admin/media/upload/" 201 8730ms
INFO "POST /ghost/api/admin/media/upload/" 201 202634ms
INFO "POST /ghost/api/admin/media/upload/" 201 6576ms
INFO "POST /ghost/api/admin/media/upload/" 201 197616ms
INFO "POST /ghost/api/admin/media/upload/" 201 5521ms
INFO "POST /ghost/api/admin/media/upload/" 201 209844ms
```

The conclusion there is no change is recommended to the current `proxy_request_buffering` directive either.

### DIFF
--- a/extensions/nginx/templates/nginx-ssl.conf
+++ b/extensions/nginx/templates/nginx-ssl.conf
@@ -22,5 +22,5 @@ server {
         allow all;
     }
 
-    client_max_body_size 50m;
+    client_max_body_size 1g;
 }


ErisDS:
Hey @markstos, as I explained on the issue, I think the proper solution here is just to increase client_max_body_size. If there's a strong need to limit uploads based on user role, that's a product concern.

I really appreciate all the research you did here, especially into the buffering config. Would you be happy to update this to just change the limit?

markstos:
@ErisDS I've forced-pushed an alternate patch raises the client_max_body_size to 1 gig to match Ghost(Pro) and better support video and other large files other self-hosters.

I checked the Nginx dogs that "g" is the right way to express Gigabytes:

https://nginx.org/en/docs/syntax.html

PR #1749, opened 2023-02-14
fix: install with proxy agent fail

Pcrab:
packageJson requires library "got" and pass agent option directly to it, but "got" checks the agent option and only allow "http", "https" or "http2" to be its keys. If pass HttpsProxyAgent without wrapping it with a correct key, finally it will throw an error.

The wrong type annotation which was fixed in package-json v8.0.0, but also changes it into an ESM only library. Ghost-Cli cannot upgrade it to fix type annotation.


KonngExplorer:
mark

PR #1769, opened 2023-07-07
🐛 Fixed not all members being exported in CLI

ronaldlangeveld:
refs https://github.com/TryGhost/Product/issues/3578

- The export API, by default is limited to 15 members.
- This sets the API endpoint to query all members via the CLI.

### DIFF
--- a/lib/tasks/import/api.js
+++ b/lib/tasks/import/api.js
@@ -150,7 +150,7 @@ async function downloadContentExport(version, url, auth, outputFile) {
 
 async function downloadMembersExport(version, url, auth, outputFile) {
     const authOpts = await getAuthOpts(version, url, auth);
-    let endpoint = '/members/upload/';
+    let endpoint = '/members/upload?limit=all';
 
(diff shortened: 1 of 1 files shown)


lib/tasks/import/api.js
```diff
@@ -150,7 +150,7 @@ async function downloadContentExport(version, url, auth, outputFile) {
 
 async function downloadMembersExport(version, url, auth, outputFile) {
     const authOpts = await getAuthOpts(version, url, auth);
-    let endpoint = '/members/upload/';
+    let endpoint = '/members/upload?limit=all';
```
daniellockyer, line 153:
```suggestion
    let endpoint = '/members/upload/?limit=all';
```

I think it's best to leave the slash otherwise it'll lead to a redirect because we enforce slashes

daniellockyer (approved):
LGTM!

We use conventional commits here, so can you squash with a message like `fix(backup): fixed exporting all members`

PR #1784, opened 2023-09-13
✨ Added support for multiple built-in themes

cmraible:
refs TryGhost/Product#3510

- Until now, Ghost has always shipped with a single built-in theme, Casper. This change adds support for Ghost to ship with multiple built-in themes.
- Updated the install and update commands to create symlinks for any theme that is shipped alongside Ghost (e.g. any theme in `ghost/core/content/themes`)
- When rolling back with `ghost update --rollback`, any symlinks that are broken in the process will be removed

### DIFF
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -131,7 +131,6 @@ class InstallCommand extends Command {
 
     defaultThemes() {
         if (fs.existsSync(path.join(process.cwd(), 'current', 'content', 'themes'))) {
-            const symlinkSync = require('symlink-or-copy').sync;
             const defaultThemes = fs.readdirSync(path.join(process.cwd(), 'current', 'content', 'themes'));
             for (const theme of defaultThemes) {
                 if (!fs.existsSync(path.join(process.cwd(), 'content', 'themes', theme))) {
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -69,7 +69,7 @@ class InstallCommand extends Command {
                     title: 'Linking latest Ghost and recording versions',
                     task: this.link.bind(this)
                 }, {
-                    title: 'Installing default themes',
+                    title: 'Linking built-in themes',
                     task: this.defaultThemes
                 }], false)
             }], {
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -130,12 +130,14 @@ class InstallCommand extends Command {
     }
 
     defaultThemes() {
-        const defaultThemes = fs.readdirSync(path.join(process.cwd(), 'current', 'content', 'themes'));
+        const currentThemesDir = path.join(process.cwd(), 'current', 'content', 'themes');
+        const contentThemesDir = path.join(process.cwd(), 'content', 'themes');
+        const defaultThemes = fs.readdirSync(currentThemesDir);
         for (const theme of defaultThemes) {
-            if (!fs.existsSync(path.join(process.cwd(), 'content', 'themes', theme))) {
+            if (!fs.existsSync(path.join(contentThemesDir, theme))) {
                 symlinkSync(
-                    path.join(process.cwd(), 'current', 'content', 'themes', theme),
-                    path.join(process.cwd(), 'content', 'themes', theme)
+                    path.join(currentThemesDir, theme),
+                    path.join(contentThemesDir, theme)
                 );
             }
         }
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -130,15 +130,13 @@ class InstallCommand extends Command {
     }
 
     defaultThemes() {
-        if (fs.existsSync(path.join(process.cwd(), 'current', 'content', 'themes'))) {
-            const defaultThemes = fs.readdirSync(path.join(process.cwd(), 'current', 'content', 'themes'));
-            for (const theme of defaultThemes) {
-                if (!fs.existsSync(path.join(process.cwd(), 'content', 'themes', theme))) {
-                    symlinkSync(
-                        path.join(process.cwd(), 'current', 'content', 'themes', theme),
-                        path.join(process.cwd(), 'content', 'themes', theme)
-                    );
-                }
+        const defaultThemes = fs.readdirSync(path.join(process.cwd(), 'current', 'content', 'themes'));
+        for (const theme of defaultThemes) {
+            if (!fs.existsSync(path.join(process.cwd(), 'content', 'themes', theme))) {
+                symlinkSync(
+                    path.join(process.cwd(), 'current', 'content', 'themes', theme),
+                    path.join(process.cwd(), 'content', 'themes', theme)
+                );
             }
         }
     }
--- a/lib/commands/install.js
+++ b/lib/commands/install.js
@@ -69,8 +69,8 @@ class InstallCommand extends Command {
                     title: 'Linking latest Ghost and recording versions',
                     task: this.link.bind(this)
                 }, {
-                    title: 'Linking latest Casper',
-                    task: this.casper
+                    title: 'Installing default themes',
+                    task: this.defaultThemes
                 }], false)
             }], {
                 argv: {...argv, version},
@@ -129,12 +129,19 @@ class InstallCommand extends Command {
         ctx.installPath = path.join(process.cwd(), 'versions', resolvedVersion); // eslint-disable-line require-atomic-updates
     }
 
-    casper() {
-        // Create a symlink to the theme from the current version
-        return symlinkSync(
-            path.join(process.cwd(), 'current', 'content', 'themes', 'casper'),
-            path.join(process.cwd(), 'content', 'themes', 'casper')
-        );
+    defaultThemes() {
+        if (fs.existsSync(path.join(process.cwd(), 'current', 'content', 'themes'))) {
+            const symlinkSync = require('symlink-or-copy').sync;
+            const defaultThemes = fs.readdirSync(path.join(process.cwd(), 'current', 'content', 'themes'));
+            for (const theme of defaultThemes) {
+                if (!fs.existsSync(path.join(process.cwd(), 'content', 'themes', theme))) {
+                    symlinkSync(
+                        path.join(process.cwd(), 'current', 'content', 'themes', theme),
+                        path.join(process.cwd(), 'content', 'themes', theme)
+                    );
+                }
+            }
+        }
     }
 
     link(ctx) {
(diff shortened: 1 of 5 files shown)


lib/commands/update.js
```diff
@@ -219,10 +219,19 @@ class UpdateCommand extends Command {
...
         symlinkSync(installPath, path.join(process.cwd(), 'current'));
 
+        // If upgrading from a version < 5.63.0, we need to symlink masthead so it is installed and available to update
+        if (semver.lt(instance.version, '5.63.0') && semver.gte(version, '5.63.0')) {
+            symlinkSync(
+                path.join(process.cwd(), 'current', 'content', 'themes', 'masthead'),
+                path.join(process.cwd(), 'content', 'themes', 'masthead')
+            );
+        }
+
```
vikaspotluri123, line 234: I think this should be done as a [migration](https://github.com/TryGhost/Ghost-CLI/blob/main/lib/migrations.js)

cmraible: Thanks! I tried this out because I agree it's a bit awkward to put this code here, but it didn't do what I need it to 😔. The migrations file you linked to are run depending on the _CLI version_ that [installed the instance of Ghost](https://github.com/TryGhost/Ghost-CLI/blob/a532d03231668f2076c0966760ff3c1e88f3b9e7/lib/utils/needed-migrations.js), and I need to create this symlink depending on the _Ghost instance version_. Basically, the migrations you're referring to are dependent on _how Ghost was installed_, rather than _which version of Ghost was installed_.

Basically, Ghost `5.62.0` doesn't include this theme, but `5.63.0+` will. We could of course check the Ghost version inside the migration and that would work in most cases (e.g. when upgrading sites on Ghost (Pro)), but you could theoretically use the latest version of the CLI (e.g. `1.24.4` once this change is released) to install an older version of Ghost (`5.62.0` or earlier) and then use the same version of the CLI to update to `5.63.0`. In this case, the migration wouldn't run because the CLI version hasn't changed and therefore the theme wouldn't be installed.

I'll probably break this block into its own task because I agree it doesn't really fit inside the `link` task, but I don't think using the existing migrations 'framework' works in this case.

vikaspotluri123: Ahh, yeah, the main issue I see is that older versions of Ghost are "vulnerable" to invalid symlinks (the stat call fails but isn't handled). If they supported invalid symlinks, you could create a symlink regardless of the Ghost version.

It would be great if Ghost's migrations could handle this, but you'd have to assume that everyone is using the CLI, and hardcode `current` into the path 🤮

Another option is to extend the migrations framework to also support `beforeInstance`. You would need to add rollback support, though. Then again, rollback support is required regardless.

lib/commands/install.js
```diff
@@ -129,12 +129,25 @@ class InstallCommand extends Command {
...
-            path.join(process.cwd(), 'content', 'themes', 'casper')
-        );
+    defaultTheme(ctx) {
+        const semver = require('semver');
+        const version = ctx.version;
+
+        // Starting with version 5.63.0, we are shipping an updated version of Casper with breaking changes under a new name
+        // If installing a version < 5.63.0, we need to symlink Casper
+        // We don't want the theme to outpace Ghost (e.g. try to use features that don't exist in Ghost yet)
+        if (semver.gte(version, '5.63.0')) {
```
vikaspotluri123, line 139: Is Casper going to be fully removed (from the bundle) in newer versions of Ghost? If so, do you need to add a task/migration to _remove_ the Casper symlink? If the theme listing service and gscan are updated to be more resilient to invalid symlinks, I guess this wouldn't be an issue.

lib/commands/install.js
```diff
@@ -129,12 +129,25 @@ class InstallCommand extends Command {
...
-        // Create a symlink to the theme from the current version
-        return symlinkSync(
-            path.join(process.cwd(), 'current', 'content', 'themes', 'casper'),
-            path.join(process.cwd(), 'content', 'themes', 'casper')
-        );
+    defaultTheme(ctx) {
+        const semver = require('semver');
+        const version = ctx.version;
+
+        // Starting with version 5.63.0, we are shipping an updated version of Casper with breaking changes under a new name
```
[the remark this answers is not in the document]
cmraible: ✅ Done

lib/commands/update.js
cmraible: Okay so I spent some time refactoring this today to address some of your concerns, and I think the result is much improved from my first attempt:

- The install command no longer checks the version number — it simply creates a symlink for _all_ themes that are shipped with Ghost. So for versions < 5.67.0, it will only link `casper` because that's the only theme that ships with Ghost, for 5.67.0+ it will link `casper` and `source`, and eventually when we remove `casper` it will only install `source`.
- The update command uses basically the same logic — it will create symlinks for any theme that installs with Ghost, so when updating to 5.67.0 it will create a symlink for `source` and in the case of a rollback, it will remove any symlinks that are broken in the process of the rollback (e.g. when rolling back from 5.67.0 it will remove the symlink to `source` because it is not present in 5.66.1).
- I also added a migration to Ghost that will be included in 5.67.0 which does _nothing_ when upgrading, but when rolling back to < 5.67.0 it will check if `source` is the active_theme and update it to `casper` if so. 

These changes paired with a migration in Ghost to set the active theme to casper _only when rolling back_ to < 5.67.0 and `source` is the current theme should address all your concerns, and I think it's a more elegant solution than manually checking the version at various points in the CLI.

lib/commands/update.js
```diff
@@ -227,6 +230,36 @@ class UpdateCommand extends Command {
         instance.version = version;
         instance.nodeVersion = process.versions.node;
     }
+
+    linkDefaultThemes({rollback}) {
+        const currentThemesDir = path.join(process.cwd(), 'current', 'content', 'themes');
+        const contentThemesDir = path.join(process.cwd(), 'content', 'themes');
```
vikaspotluri123, line 236:
```suggestion
    linkDefaultThemes({instance, rollback}) {
        const currentThemesDir = path.join(process.cwd(), 'current', 'content', 'themes');
        const contentThemesDir = path.join(instance.config.get('paths.contentPath'), 'themes');
```

lib/commands/update.js
```diff
@@ -227,6 +230,36 @@ class UpdateCommand extends Command {
...
+                for (const theme of installedThemes) {
+                    if (!fs.existsSync(path.join(contentThemesDir, theme))) {
+                        fs.rmSync(path.join(contentThemesDir, theme));
+                    }
+                }
+            }
+        } 
+
+        // ensure all default themes (e.g. themes shipped with Ghost) are symlinked to /content/themes directory
+        if (fs.existsSync(path.join(process.cwd(), 'current', 'content', 'themes'))) {
```
vikaspotluri123, line 250:
```suggestion
        if (fs.existsSync(currentThemesDir)) {
```

lib/commands/update.js
```diff
@@ -227,6 +230,36 @@ class UpdateCommand extends Command {
...
         instance.nodeVersion = process.versions.node;
     }
+
+    linkDefaultThemes({rollback}) {
+        const currentThemesDir = path.join(process.cwd(), 'current', 'content', 'themes');
+        const contentThemesDir = path.join(process.cwd(), 'content', 'themes');
+        // remove any broken symlinks caused by default themes no longer existing in previous version
+        if (rollback) {
+            if (fs.existsSync(currentThemesDir)) {
+                const installedThemes = fs.readdirSync(contentThemesDir);
```
vikaspotluri123, line 240:
```suggestion
            if (fs.existsSync(contentThemesDir)) {
                const installedThemes = fs.readdirSync(contentThemesDir);
```

Not sure if this was intentional!

lib/commands/update.js
```diff
@@ -227,6 +230,36 @@ class UpdateCommand extends Command {
...
+            const symlinkSync = require('symlink-or-copy').sync;
+            const defaultThemes = fs.readdirSync(path.join(process.cwd(), 'current', 'content', 'themes'));
+            for (const theme of defaultThemes) {
+                if (!fs.existsSync(path.join(process.cwd(), 'content', 'themes', theme))) {
+                    symlinkSync(
+                        path.join(process.cwd(), 'current', 'content', 'themes', theme),
+                        path.join(process.cwd(), 'content', 'themes', theme)
+                    );
+                }
+            }
```
vikaspotluri123, line 260: It might make sense to create a util for this since it's needed both for installing and upgrading themes

lib/commands/update.js
vikaspotluri123: Looks good overall! Left a few comments and suggestions 😊

lib/commands/update.js
cmraible: Thanks!

lib/commands/install.js
```diff
@@ -129,12 +129,19 @@ class InstallCommand extends Command {
...
     }
 
-    casper() {
-        // Create a symlink to the theme from the current version
-        return symlinkSync(
-            path.join(process.cwd(), 'current', 'content', 'themes', 'casper'),
-            path.join(process.cwd(), 'content', 'themes', 'casper')
-        );
+    defaultThemes() {
+        if (fs.existsSync(path.join(process.cwd(), 'current', 'content', 'themes'))) {
```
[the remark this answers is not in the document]
cmraible: Agreed! Removed.

lib/commands/update.js
cmraible: Ah, thank you! Fixed.

lib/commands/update.js
cmraible: TY! Fixed.

lib/commands/update.js
cmraible: Generally agree with you here, but this ended up creating a bit of a rabbithole with testing — since Ghost is shipping with the new theme tomorrow, I'm going to merge this as-is, and come back to refactor when time permits (our cleanup period starts next week).

lib/commands/update.js
```diff
@@ -92,6 +93,9 @@ class UpdateCommand extends Command {
         }, {
             title: 'Linking latest Ghost and recording versions',
             task: this.link
+        }, {
+            title: 'Installing default themes',
```
vikaspotluri123, line 97:
```suggestion
            title: 'Linking built-in themes',
```

not a critical issue, but it would align with the install task 😁

daniellockyer: Fixed in https://github.com/TryGhost/Ghost-CLI/commit/70e3be33efbadc072f54b216c5b0ef31f5362926, thanks!

Issue #1821, opened 2024-02-28
nginx and ssl setup fails due to assuming presence of certain /etc/nginx directories


20kaushik02:
### Summary

The nginx part of the setup process fails as the Ghost CLI relies on the presence of the `/etc/nginx/sites-available` folder and the SSL part fails due to the reliance on the presence of the `/etc/nginx/snippets` folder. I assume this reliance is because of how Ubuntu and some other package repositories maintain this convention in their version of nginx in their PPA.

The reliance can be seen [here on /etc/nginx/sites-available and sites-enabled](https://github.com/TryGhost/Ghost-CLI/blob/main/extensions/nginx/index.js#L112) and [here on /etc/nginx/snippets](https://github.com/TryGhost/Ghost-CLI/blob/main/extensions/nginx/index.js#L124) which throw errors

Reproduce by installing nginx from [nginx.org](https://nginx.org/en/linux_packages.html#Ubuntu) for Ubuntu, or the corresponding distro's pre-built package. Even before running `ghost setup` or `ghost install`, you can see that the two folders do not exist. Ghost-CLI does not create these directories either, but assume their presence. This can be fixed by simply creating them with `mkdir -p`.

The underlying cause is that different package repositories distribute nginx with different configuration patterns. Some have a `conf.d/*.conf` pattern (official [nginx.org](nginx.org) release), some have the `sites-available, sites-enabled` pattern (older releases that follow this Apache-like style), some have neither and have only the default `nginx.conf` file (RHEL does this, and the core nginx source code has only this as well).

Both problems are solved simply by creating those directories before running ghost, so `sudo mkdir -p /etc/nginx/sites-available /etc/nginx/sites-enabled /etc/nginx/snippets` resolves this.

### Steps to Reproduce

1. Install any version of nginx distributed without the `/etc/nginx/sites-available` and `/etc/nginx/snippets` folder. For Ubuntu, follow [these](https://nginx.org/en/linux_packages.html#Ubuntu) steps. Logs below are from nginx-1.24.0
2. Run `ghost install`, or `ghost install --no-setup` followed by `ghost setup`

### Log file

First, the nginx part:

```
ghost_server@instance-20240223-1343:~$ cat .ghost/logs/ghost-cli-debug-2024-02-28T20_04_38_506Z.log
Debug Information:
    OS: Ubuntu, v22.04.4 LTS
    Node Version: v18.19.1
    Ghost Version: 5.79.6
[...]

PR #1822, opened 2024-03-07
fix(setup): missing nginx's config and snippets directories

20kaushik02:
Fix for issue #1821. This quick fix simply creates the directories if they don't exist. Added unit test spec as well.

image


20kaushik02:
Still interested...

20kaushik02:
Still here...

Issue #711, opened 2018-04-29
local setup complains about directory permissions


mlewe:
## This issue is a

* [X] Bug Report
* [ ] Feature Request

### Summary

Using ghost-cli for a local setup (`ghost install local`) fails to start ghost again because of directory permissions.

### Steps to Reproduce

1. `ghost install local` works, starts ghost
2. `ghost stop` works
3. `ghost start` fails with the permission error given below
4. `ghost start --no-setup-linux-user` ignores directory permissions, starts ghost as expected

```
$ ghost start
Found a development config but not a production config, running in development mode instead
ℹ Ensuring user is not logged in as ghost user [skipped]
ℹ Checking if logged in user is directory owner [skipped]
✖ Checking current folder permissions
✔ Validating config
✔ Checking memory availability
One or more errors occurred.

1) SystemError

Message: The path /home/foobar/ is not readable by other users on the system.
This can cause issues with the CLI, please either make this directory readable by others or run `ghost install` in another location.

Debug Information:
    OS: Arch, vrolling
    Node Version: v6.14.1
    Ghost-CLI Version: 1.7.1
    Environment: development
    Command: 'ghost start'

Try running ghost doctor to check your system for known issues.

Please refer to https://docs.ghost.org/v1/docs/troubleshooting#section-cli-errors for troubleshooting.
```

### Technical details (for a bug report)

This is automatically output by Ghost-CLI if an error occurs, please copy & paste:

 * OS: Arch, vrolling
 * Node Version: v6.14.1
 * Ghost-CLI Version: 1.7.1
 * Environment: development
 * Command: 'ghost start'

## Bug submission checklist

Please fill out this checklist to acknowledge that you followed the requirements to submit a bug report.

- [X] Tried to find help in the forum & docs
- [X] Checked for existing issues
- [ ] Attached log file … no log file exists
- [X] Provided technical details incl. operating system

AileenCGN:
Hey @mlewe 👋 

Thanks for the bug report! I can confirm that on install, we pass this task, because the `local` argument [is given](https://github.com/TryGhost/Ghost-CLI/blob/master/lib/commands/doctor/checks/install-folder-permissions.js#L19). But for the commands `start` and `update` it's missing and we actually run this check. This is not correct, as we should pass it for local installs as well. I'll add a fix for this and it should be solved in the next release. In the meantime, you will have to use ne `--no-setup-linux-user` flag for `ghost update` and `ghost start` as a workaround.

[...]

PR #1897, opened 2024-11-10
🐛 fixed node permissions check running on local installs

vikaspotluri123:
refs #711 

- It's better to review by-commit
- Also move `isLocal` computation to the process manager rather than the doctor checks
- This can be rebased to main


acburdine (approved):
looks good to me 👍🏻

PR #1901, opened 2024-11-12
fix(install): pass agent correctly to packageJson function

### LINKED ISSUE #1899: ghost install failed with 'Cannot read properties of undefined (reading 'statusCode')' when use a proxy
An error occurred.
Message: 'Cannot read properties of undefined (reading 'statusCode')'

Debug Information:
    OS: Ubuntu, v22.04.1 LTS
    Node Version: v20.18.0
    Ghost-CLI Version: 1.26.1
    Environment: production
    Command: 'ghost install'

- issue code: 
  - https://github.com/TryGhost/Ghost-CLI/blob/v1.26.1/lib/utils/version.js#L17
  - https://github.com/TryGhost/Ghost-CLI/blob/main/lib/utils/get-proxy-agent.js#L14
- related code: https://github.com/sindresorhus/package-json/blob/v7.0.0/index.js#L57

packageJson need a agent object like `{http: httpAgent, https: httpsAgent}`, but ghost cli pass httpsAgent directly.

acburdine:
closes #1899
- at some point the syntax changed, so we need to pass the agent differently

### DIFF
--- a/lib/tasks/yarn-install.js
+++ b/lib/tasks/yarn-install.js
@@ -15,9 +15,10 @@ const getProxyAgent = require('../utils/get-proxy-agent');
 
 const subTasks = {
     async dist(ctx) {
+        const proxyAgent = getProxyAgent();
         const {dist, engines = {}} = await packageInfo('ghost', {
             version: ctx.version,
-            agent: getProxyAgent()
+            agent: proxyAgent ? {https: proxyAgent} : false
         });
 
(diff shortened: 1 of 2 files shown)


vikaspotluri123:
Aside: should we update our E2E test to use a local proxy?

acburdine:
> Aside: should we update our E2E test to use a local proxy?

hmm, possibly - i'll have to see if there's an easy way to assert in the test that traffic is actually going through the proxy

acburdine:
tested this locally and everything looks good after fixing another location 😅

PR #1945, opened 2025-04-01
nginx: remove ocsp support, its EOL at letsencrypt

csett86:
Letsencrypt will turn off OCSP responders in August 2025, so lets remove already now.

For more details see https://letsencrypt.org/2024/12/05/ending-ocsp/


acburdine (approved):

acburdine:
Thanks again for the fix here 😄 it was released in Ghost-CLI v1.28.1

PR #1951, opened 2025-04-24
🐛 Update progress icon for improved terminal compatibility 

ice9js:
A minor UI issue I discovered while using Ghost CLI within `tmux` - a popular terminal multiplexer.  
It seems that the byte-width of `hamburger` spinner icon used as the progress indicator is not well defined between terminals and locales, which in my example is causing it to unintentionally delete preceding characters.

Switching to the more common `dots` fixes the issue, but any other icon with a known and fixed width will do the trick too.

### Before

https://github.com/user-attachments/assets/552244fb-9921-4cdb-ae8c-97e6a6750edd

### After

https://github.com/user-attachments/assets/f79c8430-e132-4ea2-b546-2896746ba9e1

### DIFF
--- a/lib/ui/index.js
+++ b/lib/ui/index.js
@@ -91,7 +91,7 @@ class UI {
     run(promiseOrFunction, name, options = {}) {
         const {
             text = name,
-            spinner = 'hamburger',
+            spinner = 'dots',
             stream = this.stdout,
             clear = false,
             quiet = false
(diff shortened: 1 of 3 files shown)


acburdine (approved):
lgtm 👍🏻 we used the hamburger icon back when that more closely resembled the Ghost logo 😄

PR #1963, opened 2025-07-17
Added activitypub path to nginx

rmgpinto:
ref no-issue

- Changed nginx templates to add activitypub paths for Ghost.

### DIFF
--- a/extensions/nginx/templates/nginx-ssl.conf
+++ b/extensions/nginx/templates/nginx-ssl.conf
@@ -14,6 +14,26 @@ server {
     ssl_certificate_key <%= privkey %>;
     include <%= sslparams %>;
 
+    location ~ /.ghost/activitypub/* {
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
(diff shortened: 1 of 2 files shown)


acburdine (commented):
should we make enabling this opt-in with a flag (or maybe opt-out but on by default?)

ErisDS:
I don't think there's a need? 

The feature itself in Ghost is on-by default and having to enable it here first would add an unnecessary confusing barrier to entry.

If the feature is turned off, this rule won't get used, but having it won't really do anything.

If anyone really doesn't want it, they can delete it, which is fairly easy. I am not sure we have a good pattern for opting out in Ghost-CLI that is easier to understand than "go delete the block"?

acburdine (approved):

vikaspotluri123:
Does nginx validate the hostnames for proxy_pass directives on startup? [This forum topic](https://forum.ghost.org/t/the-following-error-occurs-nginx-host-not-found-in-upstream-ap-ghost-org/59107) reports an issue with `ap.ghost.org` after restarting their server

PR #1966, opened 2025-07-23
Added check for Python (+ setuptools), required for Node 22+

sam-lord:
Sqlite3 doesn't ship binaries for Node 22+, so it needs to be compiled instead. That relies on Python's setuptools. For Python versions <3.12 this works out-of-the-box, but for Python 3.12+ this requires the package be installed.

This PR adds a new check enabled for Node 22+, only when using SQLite (either explicitly or through a local install). The check validates that either: 
* Python <3.12 is installed
* Python 3.12+ _and_ `setuptools` are installed

### DIFF
--- a/lib/commands/doctor/checks/index.js
+++ b/lib/commands/doctor/checks/index.js
@@ -13,6 +13,7 @@ const contentFolder = require('./content-folder');
 const checkMemory = require('./check-memory');
 const binaryDeps = require('./binary-deps');
 const freeSpace = require('./free-space');
+const pythonSetuptools = require('./python-setuptools');
 
 module.exports = [
     nodeVersion,
@@ -28,5 +29,6 @@ module.exports = [
     contentFolder,
     checkMemory,
     binaryDeps,
-    freeSpace
+    freeSpace,
+    pythonSetuptools
 ];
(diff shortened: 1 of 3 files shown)


kolbma:
Can you please explain why _"Sqlite3 doesn't ship binaries for Node 22+"_?  
There is napi v3 which is compatible with Node 22.  
I've used ghost-cli v1.26.0 with sqlite and Node 22 without problems in Linux.

kolbma:
@acburdine no local installation

Could there be the relation to updated api version 10 in 22.14.0?  
Something like https://github.com/nodejs/node/issues/57118#issuecomment-2690261742

So fixed in `prebuild-install 7.1.3`, but in `yarn.lock` of ghost is used `prebuild-install 7.1.1`.  
And there would be no need to compile sqlite3 any longer.

For compiling sqlite3, next to python and setuptools, there are also some additional requirements:
* make
* g++ (C++ compiler)
* linux-headers

Without these requirements ghost-cli runs successful, but sqlite3 becomes not installed and ghost-cli hides any meaningful error why this fails. There is only some note about an optional dependency, which might be the not so optional (if you'd like to use it) dependency sqlite3.

My last ghost installation has been with nodejs 22.11.0.  
Afterwards only nodejs got updated.

Issue #1952, opened 2025-04-19
The `ghost backup` command no longer works


yannickcof:
### Issue Summary

Impossible to BACKUP (ghost backup) due to 2FA

### Steps to Reproduce

ghost backup

### Ghost Version

5.118.0

### Node.js Version

20.19.0

### How did you install Ghost?

self hosting/ Ubuntu 22.xxx

### Database type

MySQL 8

### Relevant log / error output

```shell
Response code 403 (Forbidden)
```

### Code of Conduct

- [x] I agree to be friendly and polite to people in this repository

ErisDS:
I've updated the issue title and moved it to Ghost-CLI, because _this_ problem is specific to the CLI tool. Although there ar related issues to resolve with the API in Ghost itself - the basic problem with `ghost-backup` mostly needs to be resolved her in Ghost-CLI

There are 2 authentication methods for Ghost's Admin API: session-based user authentication with a username and password, or JWT based auth using either integration tokens or staff tokens.

The `ghost backup` command is currently using User authentication, taking a username and password, which assumes the user is present. It currently doesn't have a prompt for the auth code, and anyway I'm assuming that the usecase for `ghost-backup` is something automated, or "off session" as it were, in which case user authentication isn't the right approach.

Essentially, I think the `ghost backup` command needs updating to use token auth here in Ghost CLI.

However, there is some gotchas to using token auth. 

1. Right now, staff tokens don't have enough permissions to do a DB export, and nor do _standard_ integration tokens. However, there is a special internal Ghost Backup integration that does have the right permission to do the DB backup...

2. The special Ghost Backup Integration does have permission to export the DB, but not members, which the `ghost-backup` command also does.

---

## Getting to a solution:

The main body of the work needed is to update the `ghost-backup` command to be able to take a token and use that token correctly. Ghost-CLI is usually maintained by the community, so I'd welcome help to resolve this part.

Meanwhile, I will be updating the backup integration to have enough perms, as well as revisiting permissions for staff tokens. 

---

## A potential workaround

If all you need right now is a backup of your data, that can be achieved via the API directly, without using the `ghost-backup` command. I've written a quick version of a script that will trigger a content backup, the same way `ghost-backup` does:

https://gist.github.com/ErisDS/1cc4d1a55a2f12fe31d86bb6c5d71961

Note: you need to run a query on your DB to get the id and secret from the Ghost Backup Integration.

vikaspotluri123:
[...]

PR #1967, opened 2025-07-24
feat(backup): support specifying backup staff token via env var

acburdine:
refs #1952
- add support for GHOST_CLI_STAFF_AUTH_TOKEN env var to remove the need for prompting in automated backup scenarios


vikaspotluri123 (approved):
Looks good from a diff perspective, didn't get a chance to try it out locally

acburdine:
tested it locally and in my test production instance and it worked 👍🏻

PR #1971, opened 2025-07-29
Fixed nginx config to pass .well-known to Ghost by default

sam-lord:
ref [PROD-2362](https://linear.app/ghost/issue/PROD-2362/ghost-clis-nginx-config-doesnt-work-properly-with-apghostorg)

We were passing anything matching /.well-known to Nginx, causing 404s for /ghost/.well-known/jwks.json and /.well-known/recommendations.json

I believe the only use for this was /.well-known/acme-challenge, so I've increased the specificity and only left this block in-tact for the HTTP(80) config file.

### DIFF
--- a/extensions/nginx/templates/nginx-ssl.conf
+++ b/extensions/nginx/templates/nginx-ssl.conf
@@ -44,9 +44,5 @@ server {
         add_header X-Content-Type-Options $header_content_type_options;
     }
 
-    location ~ /.well-known {
-        allow all;
-    }
-
     client_max_body_size 1g;
 }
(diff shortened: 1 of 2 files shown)


acburdine (approved):
LGTM - I'll do a double check with a test production install as well to verify 👍

sam-lord:
@acburdine Thanks for the speedy review. I just wanted to check from your recollection, were these blocks of config added specifically for ACME challenges? If there was anything else these could be used for, I don't want to cause regressions.

acburdine:
I'm pretty sure that's what they're for, yeah. I'll go back through commit history and double check.

PR #2256, opened 2026-07-30
fix(nginx): resolve ActivityPub upstream at request time

jatinder14:
## Summary
- Nginx resolves hostnames in literal \`proxy_pass\` directives at config load.
- If \`ap.ghost.org\` is briefly unresolvable (reboot / systemd-resolved restart), nginx fails to start and the site stays down.
- ActivityPub locations now use a variable upstream plus \`resolver\` so DNS happens at request time.

## Notes
- Regenerates on next \`ghost setup nginx\` / SSL setup using the CLI templates.
- Existing site configs need a regenerate (or manual edit) to pick this up.

## Test plan
- [ ] Diff generated nginx conf and confirm ActivityPub blocks use \`set \$ghost_activitypub_upstream\` + \`proxy_pass \$ghost_activitypub_upstream\`
- [ ] Confirm site still proxies \`/.ghost/activitypub/\` and webfinger/nodeinfo

Closes #2044

### DIFF
--- a/extensions/nginx/templates/nginx-ssl.conf
+++ b/extensions/nginx/templates/nginx-ssl.conf
@@ -14,6 +14,12 @@ server {
     ssl_certificate_key <%= privkey %>;
     include <%= sslparams %>;
 
+    # Resolve ap.ghost.org at request time so nginx can start when DNS is briefly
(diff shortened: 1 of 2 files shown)


acburdine:
fixed in a slightly different way in #2274 - I appreciate the initial fix here though! 😄

PR #2274, opened 2026-08-11
🐛 Fixed nginx failing to start when ap.ghost.org is unresolvable

acburdine:
closes https://github.com/TryGhost/Ghost-CLI/issues/2044

nginx resolves hostnames in literal `proxy_pass` directives when it loads its config, so a momentary DNS outage - a reboot, or systemd-resolved restarting during an unattended upgrade - made the whole site fail to start with `host not found in upstream "ap.ghost.org"`.

The ActivityPub locations now assign the upstream to a variable and declare a `resolver`, which defers the lookup to request time. A DNS outage now only affects the ActivityPub routes, and nginx starts either way.

The resolver addresses are read from /etc/resolv.conf at setup time (nginx OSS can't read it itself), so split-horizon and internal DNS keep working; public resolvers are only used if resolv.conf lists none.

Existing installs are updated by a migration, which rewrites the ActivityPub locations in place, runs `nginx -t`, and restores the original config if the result doesn't validate.


acburdine:
@coderabbitai review

macguru:
@acburdine thank you for the fix. This has been bugging me for a long time and none of my workarounds were able to fix it fully!

